Issue #31, opened 2019-09-23
[FEA] Add preview dashboard within notebook (python version)


AjayThorve:
Inline preview of how the dashboard on .show() would look like as a webapp.

**Additional context**
```python
import cudf
from cuXfilter import charts, DataFrame

df = cudf.DataFrame({'key': [0, 1, 2, 3, 4], 'val':[float(i + 10) for i in range(5)]})
cux_df = DataFrame.from_dataframe(df)

line_chart = charts.bokeh.line('key', 'val', data_points=5)
bar_chart = charts.bokeh.bar('key', 'val', data_points=5)
d = cux_df.dashboard([line_chart, bar_chart], title='Custom dashboard', data_size_widget=True)


d.preview()
```
output(image) preview of how the dashboard would look like in a full screen environment

AjayThorve:
Usage:
```python
import cudf
from cuXfilter import charts, DataFrame

df = cudf.DataFrame({'key': [0, 1, 2, 3, 4], 'val':[float(i + 10) for i in range(5)]})
cux_df = DataFrame.from_dataframe(df)

line_chart = charts.bokeh.line('key', 'val', data_points=5)
bar_chart = charts.bokeh.bar('key', 'val', data_points=5)
d = cux_df.dashboard([line_chart, bar_chart], title='Custom dashboard', data_size_widget=True)


await d.preview()
```
![image](https://user-images.githubusercontent.com/20476096/65646707-cb538f80-dfb0-11e9-8129-8baa94e2cc5c.png)

PR #38, opened 2019-09-25
Added async preview function, and updated the documentations accordingly

AjayThorve:
Use `await dashboard.preview()` to get an inline preview of how the dashboard would look layout wise, in a png format


AjayThorve:
Implements #31

PR #70, opened 2019-11-19
Fix path to conda environment

jacobtomlinson:
When installing following the [conda environment instructions](https://github.com/rapidsai/cuxfilter#2-if-installing-in-a-conda-environment) from the README the environment file referenced in the `Makefile` appears not to exist. 

This PR updates the path to what I think might be the correct environment file.


jakirkham:
Thanks all! 😄

PR #97, opened 2020-02-04
Fix gpuCI Conda builds for Python 3.6

mike-wendt:
There have been no py36 uploads for over the past month. This is a fix to correct that.


mike-wendt:
Changes were merged by auto-merger in #98

PR #117, opened 2020-03-27
[REVIEW] Let rapids-env dictate pandas version

raydouglass:


raydouglass:
> Should probably just remove numpy and cupy as well in the same vain?

Good idea

PR #122, opened 2020-03-27
[critical] bug fix which resolves rendering issue with 3d choropleth

AjayThorve:
- resolves rendering issue for choropleth 3d(deckgl)


cwharris:
How come the test was deleted?

AjayThorve:
the tests were for bokeh choropleth chart, we replaced the chart implementation to use deckgl chart, so the tests weren't related

PR #131, opened 2020-04-24
Filter fix for non aggregate charts

AjayThorve:
bug fix where generate chart is called twice which led to box-select tool filter not working on the following charts:
1. scatter
2. scatter-geo
3. datashader line
4. datashader stacked lines
5. heatmap


GPUtester:

PR #132, opened 2020-04-24
Aggregate filter accuracy fix

AjayThorve:
Aggregate charts were facing an issue where the datatiles computed weren't accurate for range selections(were missing zero value indexes)
- fixed updating chart data as per accurate indices
- fixed sending incorrect cuDF to aggregate data_tile compute
- updated tests


GPUtester:

PR #135, opened 2020-04-28
making non-aggregate query stateless

AjayThorve:
Non-aggregate chart type queries were current state dependent. This PR converts non-aggregate query to being stateless, a step closer to multi-user dashboard experience.


GPUtester:

PR #159, opened 2020-06-11
[WIP] Layouts Refactor

bryevdv:
Hi @AjayThorve I've tried to split up and arrange the commits in a fairly atomic way, but the new `layouts` module is down to ~300 lines so it might be useful to just refer to the entire new file as well. I tested manually with the available notebooks and with the existing unit tests (more about tests below)

#### Factor out predicates

I factored the detection of widgets out into a function `is_widget`. My own preferences here are aggressive. It's not an exaggeration to say that once a predicate has an "and" or "or" I'm about ready to pull it into its own function. 

#### Use built-in `enumerate`

Almost always a win to use `enumerate` over manually fiddling with counters

#### Split loops

I know the inclination here is probably towards loop fusion :) But in this case there are only ever single digit numbers of things to loop over, so the logic can be more clearly expressed by not doing multiple things in one big loop. In this case: theme application, widget filtering, and chart templating all happened in one loop. This PR splits out those tasks into their own functions (which could be individually tested), which also affords the next step:

#### NVI pattern

The addition of the base class `_LayoutBase` was to implement the [Non-virtual Interface Pattern](https://en.wikipedia.org/wiki/Non-virtual_interface_pattern). This is a really useful pattern when there are fixed steps that surround some kind of varying steps, as was the case here. Specifically, the theme application, widget handling, and "padding" step are all the same. What varies is just the aspect ratio handling, so that is all that is pulled in to the `_process_plots` method on subclasses. 

It would have also been possible to accomplish this with a decorator but it would be a fairly complicated decorator, so I went with the base class in this case. 

All of the above was somewhat aggressive "factoring out common code" but like to also always remember a main reason it is good to factor out common code actually involves all the code paths that are *not* common, i.e. to make them easier to see and compare 

#### Collapse identical branches

Many of the aspect-ratio computations were identical, so they could be collapsed in to branches that handled any of the relevant plots via `if i in [...]`

#### Tests

There weren't any existing tests. I am happy to take a poke at adding some, in this PR or in follow up. It would be helpful if there is any existing easy way to manually (e.g. an informal test script) it would be nice to go through that. I tried to make "safe" transformations but there is always the possibility of typos, etc. 

#### Factor Templates

The templates themselves also have a fair amount of duplication. This could be factored out with either text subsitution, or perhaps preferably) actually Jinja template inheritance 

#### More Data-driven

In the ideal case I think it would be possible to define a declarative description for each layout that allows all the following to be derived from a single "source of truth":
* the template itself
* the aspect ratios 

If that was done then even `_process_plots` could be generalized and most of `layouts.py` deleted. That might getting too far in to yak-shaving territory though

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# cuXfilter 0.15.0 (unreleased)
+ 
+## Improvements
+- PR #157 Layouts Refactor
+
 # cuXfilter 0.14.0 (03 Jun 2020)
 
 ## New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## New Features
 
 ## Improvements
-- PR #157 Layouts Refactor
 - PR #158 Add docs build script
+- PR #159 Layouts Refactor
 
 ## Bug Fixes
 
--- a/python/cuxfilter/layouts/layouts.py
+++ b/python/cuxfilter/layouts/layouts.py
@@ -82,9 +82,8 @@ class Layout0(_LayoutBase):
         charts = [x for x in charts.values() if not is_widget(x)]
 
         for i, chart in enumerate(charts, 1):
-            if i == 1:
-                chart.width = 1600
-                chart.height = int(round(90 * 1.0)) * 10
+            chart.width = 1600
+            chart.height = 900
             chart.chart.sizing_mode = "scale_both"
             tmpl.add_panel("chart{i}", chart.view())
 
@@ -106,10 +105,10 @@ class Layout1(_LayoutBase):
         for i, chart in enumerate(charts, 1):
             if i == 1:
                 chart.width = 1600
-                chart.height = int(round(90 * 0.66)) * 10
+                chart.height = 600
             elif i == 2:
                 chart.width = 1600
-                chart.height = int(round(90 * 0.33)) * 10
+                chart.height = 300
             chart.chart.sizing_mode = "scale_both"
             tmpl.add_panel("chart{i}", chart.view())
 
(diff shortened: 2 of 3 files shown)


GPUtester:

AjayThorve:
looks good to me, all good suggestions.

tests: we should certainly add more tests. Other sections are also missing unit tests, and there are no integration tests currently, might as well be worth it to create a separate PR for tests altogether


Factor Templates:
I like the idea of jinja template inheritance, the navbar section is redundant across, and container charts could be part of a for loop


More data-driven:
We could do that, the thought process behind current way was to make sure cuXfilter is opinionated, and have fixed layouts - https://docs.rapids.ai/api/cuxfilter/stable/layouts/Layouts.html

But I could see how suggested approach makes sense.

bryevdv:
@AjayThorve I think opinionated is fine :) The data-driven suggestion still comes from primarily a maintainability POV. What is the next step to get this PR green, just a CHANGELOG entry at some point? (And who is generally responsible for pushing the merge button after green+approval?)

AjayThorve:
Changelog entry would do it, would need to add a new section of 0.15 release. That would trigger the CI, and then you should be able to merge it

exactlyallan:
This is looking great @bryevdv

bryevdv:
@AjayThorve IIRC there is a changeling started in another WIP PR? Do you want me to add one this PR ahead of that one? Or wait for the other and then merge/rebase on top of it?

AjayThorve:
@bryevdv I think ahead of that one should be fine, since all changes are referred to by the PR numbers, any conflicts could be resolved pre-merge incase there are any.

PR #162, opened 2020-06-19
Dashboard and templates cleanup and tests

bryevdv:
@AjayThorve Here is some more light refactoring/cleanup. I think most of the commits are self-contained / self-explanatory but it's also a bit of a hodge-podge, since its just things I noticed while reading to learn the codebase, so a few comments:

* ***De-duplication*** Much of the PR is just refactoring things that are duplicated to not be duplicated.

* ***Optional values*** I changed a couple of places that used "" as an missing optional. None is more idiomatic and plays better with MyPy e.g. `Optional[str]`, or alternatively just using an actual real default value. 

* ***Unnecessary properties*** My thoughts on these have definitely evolved. :) I also used to favor them but after spending some time with some Python core devs at MS, I have come around to the notion that properties like the ones removed here are generally unnecessary and best omitted (Steve Dower goes so far as to refer to them as anti-pattern!)

I also added some simple tests to validate and preserve that the layout templates embed what is expected and necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 ## Improvements
 - PR #158 Add docs build script
-- PR #160 Install dependencies via meta packages
 - PR #159 Layouts Refactor
+- PR #160 Install dependencies via meta packages
+- PR #162 Dashboard and templates cleanup and tests
 
 ## Bug Fixes
 - PR #161 fixed layouts bugs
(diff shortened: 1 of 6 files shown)


GPUtester:

AjayThorve:
Everything looks good to me. 

The point about the unncessary properties does make sense, and the ones removed were actually not required(the initial aim was to type check those properties, but now that I think of it, they are fairly intuitive). Thanks !

PR #168, opened 2020-07-16
Legend support and refactor [Review]

AjayThorve:
This PR includes some code refactor, majorly removing redundant classes:
- Merged scatter_geo and scatter api. Now unified to scatter (with tile_provider(underlying maps) providing geo chart support)
- added legends to following non_aggregate charts (for linear and log pixel_shade_type - `datashader.transfer_function.shade.how` value):
1. scatter
2. heatmap
3. stacked_lines
4. graphs (nodes)

Screenshots(in above order):

image

image

image

image


GPUtester:

GPUtester:

bryevdv:
What's going on with that colorbar position the heatmap? I am surprised to see it so far offset from the plot.

python/cuxfilter/charts/datashader/datashader.py
```diff
@@ -128,13 +13,16 @@ def scatter(
     aggregate_fn="count",
     point_size=1,
     point_shape="circle",
-    pixel_shade_type="eq_hist",
+    pixel_shade_type="linear",
```
bryevdv, line 16: Why the change from `eq_hist`?

python/cuxfilter/charts/datashader/datashader.py
```diff
@@ -474,10 +389,17 @@ def heatmap(
...
+        tile_provider=None,
+        title=title,
+        timeout=timeout,
+        legend=legend,
+        legend_position=legend_position,
         **library_specific_params,
     )
+    # since it's built using the scatter api,
+    # changing the chart_type after the chart is created
+    p.chart_type = "heatmap"
```
bryevdv, line 401: If `chart_type` being set by `Scatter`? If `chart_type` is some classification that can't be reliably derived automatically from the implementation alone (e.g. this result is made with a `Scatter but is a heatmap for whatever purpose `chart_type` is used for) then I'd suggest not setting it automatically at all, and just being explicit about the chart type everywhere it needs to be set in the lowercase methods here like `heatmap`. `line` etc. That way it's clear where the responsibility for setting this value lies and there  aren't overrides in some places  but not others (so then comments like this are also not needed)

python/cuxfilter/charts/datashader/datashader.py
AjayThorve: So legends are not yet supported in eq_hist, I wanted to make sure they are rendered in the default examples. Or should it be explicitly mentioned in the docs like - `if legends are a requirement, set pixel_shade_type to 'linear'/'log'`?

bryevdv: Oh right, the work for eqhist colorbars is still currently ongoing, I forgot 😞 Well that's annoying. 

I think there should be a note, regardless, since it will help set user expectations and be something that can be pointed to in case of questions.

For the examples can we simply be explicit to set `"linear"` for now (to get a colorbar) but leave the default as eq_hist? Suggest this since:

* I think `eq_hist` is the better default, current colorbar limitations aside
* It avoids having to change an API default value (twice)

python/cuxfilter/charts/datashader/datashader.py
bryevdv: If it is a property *on* the result, then `heatmap` could pass it as a configuration to `Scatter(...)` to set in its initializer (but I'd just suggest being explicit about passing the value from "outside" Scatter)

python/cuxfilter/charts/datashader/datashader.py
AjayThorve: Yep, that makes more sense

AjayThorve:
> What's going on with that colorbar position the heatmap? I am surprised to see it so far offset from the plot.

@bryevdv fixed this, was occurring due to legend layout being added more than once on top of each other.

image

bryevdv:
@AjayThorve looks much bette! I think the colorbar title overlap with the toolbar is a small bug (colorbar titles are a recent addition and I guess no one thought of this interaction possibility at the time) can you file an bug on the Bokeh tracker?

AjayThorve:
> @AjayThorve looks much better! I think the colorbar title overlap with the toolbar is a small bug (colorbar titles are a recent addition and I guess no one thought of this interaction possibility at the time) can you file an bug on the Bokeh tracker?

@bryevdv seems like there was an issue created, but it's closed now for some reason https://github.com/bokeh/bokeh/issues/5095

python/cuxfilter/charts/datashader/datashader.py
AjayThorve: fixed this, all chart_types are now explicitly set outside the base classes

bryevdv:
@AjayThorve That issue is from a couple of years before titles on colorbars existed at all. It looks like it was closed as a dupe of another issue generally requesting colorbar titles. But now that colorbar titles do exist, they seem to need some slight improvements, so I do think a new issue here is appropriate.

AjayThorve:
@bryevdv I created a Feature request here - https://github.com/bokeh/bokeh/issues/10326

PR #169, opened 2020-07-30
Graph updates

AjayThorve:
Two major features:
- Added edge-curving + bundling inspired by ->https://github.com/rapidsai/rapids-js/blob/494c2caa8f5ffda323c132fd74d54f7dc10c34ad/modules/demo/graph/python/shaping.py#L39. Since instanced rendering is not available in datashader, had to write a kernel to divide an edge into 100 sub-edges and curving it as per the control points. Although that results in 100x larger edge datasets, so it may throw OOM errors for larger datasets and directly_connected_edges are recommended for those use-cases.

- Added Node-neighbor inspect widget, which is toggleable.


GPUtester:

AjayThorve:
rerun tests

AjayThorve:
merging this, as the failures are related to numba+datashader compatibility issues, which is being tracked here - https://github.com/rapidsai/ops/issues/1097

PR #171, opened 2020-08-06
Chart fixes - critical

AjayThorve:
fix issues related to rendering deck charts


GPUtester:

AjayThorve:
rerun tests

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -7,6 +7,20 @@ class BaseAggregateChart(BaseChart):
...
+    ):
+        """
+        source_x: current_source_x, np.array()
+        source_y: current_source_y, np.array()
+        update_data_x: updated_data_x, np.array()
+        update_data_y: updated_data_x, np.array()
+        """
+        result_array = np.zeros(shape=(int(source_x.max()),))
+        # -1 for 0-based indexing
+        np.put(result_array, update_data_x.astype("int") - 1, update_data_y)
```
bryevdv, line 21: nit: this (and others) could just be `astype(int)` which is slightly preferable over magic strings

python/cuxfilter/tests/assets/test_gpu_histogram.py
```diff
@@ -8,7 +8,54 @@
...
+                        50,
+                        50,
+                        50,
+                        100,
+                    ],
+                ]
+            ),
+        ),
+    ],
+)
```
bryevdv, line 57: nit: sometimes black formatting is alot :) This might collapse a bit if the lists are pulled out in to variables to pass to the decorator

python/cuxfilter/charts/datashader/plots.py
```diff
@@ -680,7 +684,6 @@ def generate_chart(self):
             y_range=self.y_range,
             width=self.width,
             height=self.height,
-            output_backend="webgl",
```
bryevdv, line 683: just curious the motivation to remove this?

python/cuxfilter/charts/datashader/custom_extensions/graph_inspect_widget.py
```diff
@@ -0,0 +1,58 @@
...
+          this.model._active = this.model.properties.active.spec.value
+      })
+  }
+}
+
+export namespace CustomInspectTool {
+  export type Attrs = p.AttrsOf<Props>
+
+  export type Props = InspectTool.Props & {
+    _active: p.Property<Boolean>
```
bryevdv, line 25: Interesting, I guess this is this because `active` is not exposed? Now that `readonly` properties area possible I think we could make this accessible on the Python side. Please feel free to open an issue!

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
```diff
@@ -109,6 +111,8 @@ def selection_callback(xmin, xmax, ymin, ymax):
             temp_data = dashboard_cls._query(
                 dashboard_cls._generate_query_str()
             )
+            if len(temp_data) == 0:
+                temp_data = cudf.DataFrame({k: cp.nan for k in self.source})
```
bryevdv, line 115: Would it be possible to handle zero-len inside `_reload_charts` so that the caller does not have to care about this special case?

python/cuxfilter/tests/charts/core/test_core_graph.py
```diff
@@ -61,25 +62,52 @@ def test_get_selection_geometry_callback(self):
...
+        dashboard = DashBoard(dataframe=DataFrame.load_graph((nodes, edges)))
+
         bg = BaseGraph()
-        bg.node_x = "a"
-        bg.node_y = "b"
         bg.chart_type = "temp"
+        bg.nodes = nodes
+        bg.edges = edges
+        bg.inspect_neighbors = inspect_neighbors
+        print(bg.inspect_neighbors._active)
```
bryevdv, line 98: stray print in test

bryevdv:
minor comments otherwise LGTM

python/cuxfilter/charts/datashader/plots.py
AjayThorve: So initially there was an automatic switch in rendering the points, nodes<=10_000 used bokeh rendering, which is where the output_backend="webgl" was used, but there was a lot of difference between the datashader based point rendering and bokeh based point/graph rendering, one of the things being circle_radius scales. To avoid confusion, I thought its best to have bokeh based graph rendering as a separate chart(will have to add it in 0.16)

python/cuxfilter/charts/datashader/custom_extensions/graph_inspect_widget.py
AjayThorve: ya, that was precisely the reason. I'll raise on issue, thank you :)

python/cuxfilter/tests/assets/test_gpu_histogram.py
AjayThorve: fixed

python/cuxfilter/charts/core/aggregate/core_aggregate.py
AjayThorve: fixed

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
AjayThorve: fixed this. Since this is only required for non_aggregate(datashader charts), moved the DF check over to datashader reload_chart for the charts requiring it (datashader - non_aggregate, graph, stacked_lines)

python/cuxfilter/tests/charts/core/test_core_graph.py
AjayThorve: fixed

PR #177, opened 2020-08-19
add lasso selection support

bryevdv:
cc @AjayThorve 

I reworked my original approach to this. At first, I had modified `get_selection_geometry_callback` to be `get_selection_geometry_callbackS` (plural) and had it return a list of callbacks to downstream consumers. This preserves the flexibility of those intermediate `temp_callback` wrappers. However, this broke quite a bit in testing and elsewhere and I was not quite confident in some of the chain of changes I had to start making. 

The current PR pushes the event itself upstream and to the `selection_callback` directly (and eliminating the intermediate `temp_callback`  functions). Then the `selection_callback` does the switching between rect and lasso events. This significantly limits the scope of changes, at the cost of some flexibility in principle. However I note that all of the current `temp_callback`s were identical, so this flexibility was not used in practice. 

I'd advocate for this approach for now, and if circumstances in the future actually warrant, moving to a more generic approach. 

I did test this with the NYC notebook but any guidance around testing (or the code itself) would be appreciated.

### DIFF
--- a/conda/environments/cuxfilter_dev_cuda10.0.yml
+++ b/conda/environments/cuxfilter_dev_cuda10.0.yml
@@ -7,6 +7,7 @@ channels:
 - anaconda
 dependencies:
 - cupy>=6*
+- cuspatial>=0.15
 - python>=3.6,<3.8
 - cudf=0.16
 - dask>=2.12.0
@@ -25,6 +26,7 @@ dependencies:
 - nodejs
 - recommonmark
 - pytest
+- mock
 - sphinx
 - sphinx_rtd_theme
 - jupyter_sphinx
--- a/conda/environments/cuxfilter_dev_cuda10.1.yml
+++ b/conda/environments/cuxfilter_dev_cuda10.1.yml
@@ -7,6 +7,7 @@ channels:
 - anaconda
 dependencies:
 - cupy>=6*
+- cuspatial>=0.15
 - python>=3.6,<3.8
 - cudf=0.16
 - dask>=2.12.0
@@ -25,6 +26,7 @@ dependencies:
 - nodejs
 - recommonmark
 - pytest
+- mock
 - sphinx
 - sphinx_rtd_theme
 - jupyter_sphinx
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # cuXfilter 0.16.0 (Date TBD)
 
 ## New Features
+- PR #177 Add support for lasso selections
 
 ## Improvements
 - PR #191 Update doc build script for CI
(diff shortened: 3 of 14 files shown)


GPUtester:

conda/environments/cuxfilter_dev_cuda10.0.yml
```diff
@@ -7,6 +7,7 @@ channels:
 - anaconda
 dependencies:
 - cupy>=6*
+- cuspatial>=0.15
```
mike-wendt, line 10: This file should be renamed and changed to CUDA 11.0 as we no longer support CUDA 10.0

mike-wendt (changes requested):
I see you're adding

conda/environments/cuxfilter_dev_cuda10.1.yml
```diff
@@ -24,6 +25,7 @@ dependencies:
 - nodejs
 - recommonmark
 - pytest
+- mock
```
mike-wendt, line 28: Do you need `mock` for conda builds or GPU testing? If so, you need to add it here as well - https://github.com/rapidsai/integration/blob/branch-0.16/conda/recipes/rapids-build-env/meta.yaml

bryevdv: @mike-wendt it's only used in a basic (non-GPU) unit testing, it is not needed in the release package or at runtime. So I think this is sufficient? But please lmk if I have misunderstood anything

AjayThorve:
@bryevdv, for the CI to pass, we might have to add cuspatial to the gpu ci containers env for cuxfilter, as we did for cugraph -> https://github.com/rapidsai/cuxfilter/blob/branch-0.16/ci/gpu/build.sh#L61

AjayThorve (changes requested):
LGTM, just one thing that's missing is adding `LassoSelectTool()` to `class Graph` in `plots.py`

conda/environments/cuxfilter_dev_cuda10.1.yml
bryevdv: seems to have passed w/o adding anything (cc @AjayThorve )

AjayThorve (approved):

mike-wendt (approved):
Main changes have been made. My last concern is making sure that the unit tests you mention for `mock` are not triggered or run by gpuCI at any stage (conda builds or testing). The reason being is because we do not have `mock` installed in the images or conda environment for it to be used.

bryevdv:
@mike-wendt would you like a PR to add it to https://github.com/rapidsai/integration/blob/branch-0.16/conda/recipes/rapids-build-env/meta.yaml just for good measure? It's a fairly common test dependency so maybe just good to have in any case

bryevdv:
@mike-wendt actually looking at the logs it appears that the reason the tests passed that that something else is implicitly pulling in mock in the conda install already. But probably a good idea to make explicit in the recipe?

PR #192, opened 2020-09-18
Minor Fixes and drop_duplicates attr for view_dataframe()

AjayThorve:
Bug fixes
- fixes issues with non_aggregate charts having permanent inplace querying and query_by_indices

Feature Additions:
- Added drop_duplicates for view_dataframe chart type
- Optimized graph querying logic

### DIFF
--- a/python/cuxfilter/charts/core/non_aggregate/core_graph.py
+++ b/python/cuxfilter/charts/core/non_aggregate/core_graph.py
@@ -221,14 +221,13 @@ class BaseGraph(BaseChart):
                 ys,
             )
             nodes = self.source[indices.selection]
+            edges = None
 
             if self.inspect_neighbors._active:
                 node_ids = nodes[self.node_id]
                 nodes, edges = self.query_graph(
                     node_ids, self.nodes, self.edges
                 )
-            else:
-                edges = None
 
             # reload all charts with new queried data (cudf.DataFrame only)
             dashboard_cls._reload_charts(data=nodes, ignore_cols=[self.name])
@@ -256,15 +255,15 @@ class BaseGraph(BaseChart):
             )
 
             dashboard_cls._query_str_dict[self.name] = query
+
             nodes = dashboard_cls._query(dashboard_cls._generate_query_str())
+            edges = None
 
             if self.inspect_neighbors._active:
                 node_ids = nodes[self.node_id]
                 nodes, edges = self.query_graph(
                     node_ids, self.nodes, self.edges
                 )
-            else:
-                edges = None
 
             # reload all charts with new queried data (cudf.DataFrame only)
             dashboard_cls._reload_charts(data=nodes, ignore_cols=[self.name])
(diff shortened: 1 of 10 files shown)


AjayThorve:
rerun tests

python/cuxfilter/charts/core/non_aggregate/core_graph.py
```diff
@@ -254,19 +256,22 @@ def box_callback(xmin, xmax, ymin, ymax):
...
+            nodes = dashboard_cls._query(dashboard_cls._generate_query_str())
 
-            nodes, edges = self.query_graph(node_ids, self.nodes, self.edges)
+            if self.inspect_neighbors._active:
+                node_ids = nodes[self.node_id]
+                nodes, edges = self.query_graph(
+                    node_ids, self.nodes, self.edges
+                )
+            else:
+                edges = None
```
bryevdv, line 267: Purely a personal style pref but I think it would be slightly clearer as
```
# set both nodes and edges to initial values together in one place
nodes = dashboard_cls._query(dashboard_cls._generate_query_str())
edges = None

# conditionally update both nodes and edges only in one place (no else: needed)
if self.inspect_neighbors._active:
    node_ids = nodes[self.node_id]
    nodes, edges = self.query_graph(node_ids, self.nodes, self.edges)
```

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
```diff
@@ -271,23 +271,23 @@ def query_chart_by_indices(
...
             final_query = active_chart.x + " in (" + new_indices_str + ")"
             if len(query) > 0:
                 final_query += " and " + query
-            self.reload_chart(
-                self.source.query(final_query), False,
-            )
+
+        self.reload_chart(
+            self.source.query(final_query)
+            if len(final_query) > 0
```
bryevdv, line 287: Would there be downsides to letting `source.query` accept a zero length query?

bryevdv (approved):

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
AjayThorve: That results in a QuerySyntaxError
image

python/cuxfilter/charts/core/non_aggregate/core_graph.py
AjayThorve: fixed, thanks

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
bryevdv: Ah, I didn't realize `self.source` was not some wrapper we created. :+1:

bryevdv (approved):

exactlyallan (approved):

PR #194, opened 2020-10-02
updated show and app methods for jupyterhub

AjayThorve:
- Added `service_proxy` parameter to `cuxfilter.dashboard.show` and `cuxfilter.dashboard.app`, setting it to 'jupyterhub', will make sure cuxfilter dashboard works in a jupyterhub/mybinder environment.

- inspired by the bokeh implementation - http://docs.bokeh.org/en/latest/docs/user_guide/jupyter.html#jupyterhub

### DIFF
--- a/python/cuxfilter/dashboard.py
+++ b/python/cuxfilter/dashboard.py
@@ -2,9 +2,10 @@ from typing import Dict, Type
 import bokeh.embed.util as u
 import panel as pn
 import uuid
-from panel.io.server import _origin_url, get_server
+from panel.io.server import get_server
 from bokeh.embed import server_document
-import re
+import os
+import urllib
 
 from .charts.core.core_chart import BaseChart
 from .datatile import DataTile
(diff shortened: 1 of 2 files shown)


python/cuxfilter/dashboard.py
```diff
@@ -42,26 +63,25 @@ def _create_app(panel_obj, notebook_url=None, port=0):
...
 
-    url = _create_dashboard_url(notebook_url, port)
-
     script = server_document(url, resources=None)
 
     publish_display_data(
         {HTML_MIME: script, EXEC_MIME: ""},
         metadata={EXEC_MIME: {"server_id": server_id}},
     )
+    print(server_id)
```
bryevdv, line 84: spurious print?

python/cuxfilter/dashboard.py
```diff
@@ -416,6 +441,9 @@ def app(self, notebook_url=DEFAULT_NOTEBOOK_URL, port: int = 0):
             Default is random open port. Recommended to set this value if
             running jupyter remotely and only few ports are exposed.
 
+        environment: str, optional, default jupyter-notebook,
+            options: jupyter-notebook, jupyterlab, jupyterhub
```
bryevdv, line 445: Is there a better parameter name than "environment"? That word already has lots of other uses. Some ideas
```
jupyter = classic | lab | hub
notebook = classic | jupyterlab | jupyterhub
```
Or 
```
service_proxy = none | jupyterlab
```
?

AjayThorve: ah, I agree with you on this, so technically, we only have two scenarios, adding the prefix for jupyterhub, or not adding it for jupyterlab/classic, given that, I think should work:

```python
service_proxy = none(for classic/lab) | jupyterhub
```
does this seem fine?

python/cuxfilter/dashboard.py
AjayThorve: fixed

python/cuxfilter/dashboard.py
AjayThorve: replaced with service_proxy, can you have a look and see if it looks good?

exactlyallan (approved):

bryevdv:
@AjayThorve Sorry I did not see your question earlier. A side effect of having things use my existing `bryan@bokeh.org`  GitHub email address, is that I only get notifs on my personal laptop, not on my work laptop. I will have to try and see what different  accommodations I can make so that I do get notifs. Maybe there is a way to route by repo or org to a different address.

Issue #35, opened 2019-09-24
[BUG] Add support for dashboarding dataframes with string columns


AjayThorve:
**Describe the bug**
When using dataframes with columns of type string, I get the following bug -> `AttributeError: 'StringColumn' object has no attribute 'min'`

**Steps/Code to reproduce bug**
```python
from cuXfilter import charts, DataFrame
from cuXfilter.layouts import layout_1, layout_2
import cudf

df = cudf.DataFrame({'key': ['one', 'two', 'three', 'four', 'five'],'val':[float(i + 10) for i in range(5)]})

#create cuXfilter DataFrame
cux_df = DataFrame.from_dataframe(df)
line_chart = charts.bokeh.line('key', 'val', data_points=5)
bar_chart = charts.bokeh.bar('key', 'val', data_points=5)

charts = [line_chart, bar_chart]

#create the dashboard object
d = cux_df.dashboard(charts, title='Custom dashboard', data_size_widget=True)

#display the dashboard as a web-app
d.app()
```
Stack trace
```python 
AttributeError                            Traceback (most recent call last)
<ipython-input-2-4916d3157f84> in <module>
     13 
     14 #create the dashboard object
---> 15 d = cux_df.dashboard(charts, title='Custom dashboard', data_size_widget=True)
     16 
     17 #display the dashboard as a web-app

~/cuxfilter-design/cuXfilter/dataframe.py in dashboard(self, charts, layout, title, data_size_widget, warnings)
    117 
    118         """
--> 119         return DashBoard(charts, self.data, layout, title, data_size_widget, warnings)

~/cuxfilter-design/cuXfilter/dashboard.py in __init__(self, charts, data, layout, title, data_size_widget, warnings)
     75             for chart in charts:
     76                 self._charts[chart.name] = chart
---> 77                 chart.initiate_chart(self)
     78 
     79         self._title = title

~/cuxfilter-design/cuXfilter/charts/core/aggregate/core_aggregate_line.py in initiate_chart(self, dashboard_cls)
     77 
     78         '''
---> 79         self.min_value = dashboard_cls._data[self.x].min()
     80         self.max_value = dashboard_cls._data[self.x].max()
     81         if self.stride is None:

~/anaconda3/envs/cudf_0.8/lib/python3.7/site-packages/cudf/dataframe/series.py in min(self, axis, skipna, dtype)
   1708         """
   1709         assert axis in (None, 0) and skipna is True
[...]

PR #208, opened 2020-10-22
[Review] New dtypes - datetime

### LINKED ISSUE #204: [BUG] timeseries datetime64 not supported - invalid type promotion
**Describe the bug**
A `datetime64[ns]` type for a timeseries column results in TypeError: invalid type promotion when trying to plot a cuxfilter plot.

**Steps/Code to reproduce bug**

```
import pandas as pd
import cuxfilter
import cudf
from cuxfilter import DataFrame, themes, layouts, charts
from cuxfilter.assets.custom_tiles import get_provider, Vendors

df_minimal = pd.DataFrame({'foo': [1,2,3], 'time':['2020-01-01', '2020-01-10 01', '2020-02-20 22']})
df_minimal['time'] = pd.to_datetime(df_minimal['time'])

cm_cu = cudf.from_pandas(df_minimal)
cux_df = cuxfilter.DataFrame.from_dataframe(cm_cu)

chart2 = charts.line('time')
charts_list = [chart2]

d = cux_df.dashboard(charts_list, title='time series', layout=layouts.feature_and_base, theme=themes.rapids, data_size_widget=True)
```

**Expected behavior**
the datatype should be supported as this is supported by cudf and fundamentally the normal way of handling time also in pandas

**Environment details (please complete the following information):**
 - Environment location: bare metal
 - Method of cuxfilter install: conda, version 0.15

AjayThorve:
Work in Progress PR - aims to fix #204, #35 (partial - no interaction)

- Adds support for new dtype - datetime for all chart types except choropleths
- Added new chart widget type - DateRangeSlider
- Some refactor - merged BaseLine and BaseBar to BaseAggregate

Pending:
- add new tests for cuxfilter.assets.datetime
- update docs & notebooks with new examples for datetime axes


GPUtester:

bryevdv (changes requested):

bryevdv:
@AjayThorve This is a hefty PR!  left a first round of comments

AjayThorve:
> @AjayThorve This is a hefty PR!  left a first round of comments 
> 

@bryevdv apologies for the huge PR, take your time, aiming for 0.17 so no hurries.

python/cuxfilter/charts/core/non_aggregate/core_line.py
```diff
@@ -134,33 +167,37 @@ def add_range_slider_filter(self, dashboard_cls):
...
+                **{"width": self.width},
+                sizing_mode="scale_width",
+            )
+        else:
+            self.filter_widget = pn.widgets.RangeSlider(
+                start=self.min_value,
+                end=self.max_value,
+                value=(self.min_value, self.max_value),
+                step=self.stride,
+                **{"width": self.width},
```
bryevdv, line 184: I don't see how this is any different from just passing `width=`
```
In [1]: def foo(a, b, c, d): 
   ...:     print(a, b, c, d) 
   ...:                                                                                                                                                                                                                                         

In [2]: foo(a=1, b=2, c=3, d=4)                                                                                                                                                                                                                 
1 2 3 4

In [3]: foo(a=1, b=2, **{"c": 3}, d=4)                                                                                                                                                                                                          
1 2 3 4
```

python/cuxfilter/assets/datetime.py
```diff
@@ -0,0 +1,123 @@
...
+    return True
+
+
+def to_int64_if_datetime(dates, _type):
+    """
+    Description:
+        convert to int64 for input to datashader based plots
+    -------------------------------------------
+    Input:
+        col: cudf.Series | list | tuple
```
bryevdv, line 105: parameter is `dates` not `col`

python/cuxfilter/assets/datetime.py
```diff
@@ -0,0 +1,123 @@
...
+    if _type in CUDF_DATETIME_TYPES:
+        if type(dates) in [list, tuple]:
+            # compute date seconds factor
+            dt_s_factor = get_dt_unit_factor(dates[0], _type)
+            return (np.array(dates).astype("int64")) * dt_s_factor
+        elif type(dates) == cudf.Series and check_series_for_nan(dates):
+            # compute date seconds factor
+            dt_s_factor = get_dt_unit_factor(dates.iloc[0], _type)
+            return (dates.astype("int64")) * dt_s_factor
+    return dates
```
bryevdv, line 117: This comment does not have any specific actionable suggestion. I wish I had more fully formed thoughts, but I wanted to at least just jot down a consideration for the future. One thing that troubles me a bit is how dtypes are stored separately and redundantly. Things are either arrays (carry their own dtype) or not (no dtype) and storing this again only affords a chance for things to become out of sync. I guess instead of things roughly like
```
to_int64_if_datetime(event.geometry["x"], self.x_dtype)
```
I'd be happier to see more like
```
self.to_xaxis_type(event.geometry["x"])
```
where `to_xaxis_type` can inspect `self.source[self.x].dtype`  instead of requiring the dtype to actually be stored so it can be passed to some outside function (if a *convenience* is useful, `self.x_dtype` could be a *computed property*, instead of actually storing it as it does now). 

I know this oversimplifies things e.g dtypes are stored on charts and only reference the df on some *other* objects 
```
self.x_dtype = dashboard_cls._cuxfilter_df.data[self.x].dtype
```
but to me this only increases the likelihood that some future code changes could result in things becoming out of sync. Maybe a computed read-only property for `self.x_dtype` would be a potential quick improvement.

python/cuxfilter/charts/core/non_aggregate/core_line.py
```diff
@@ -68,22 +75,13 @@ def __init__(
...
-        -------------------------------------------
-        Input:
-        data: cudf DataFrame
-        -------------------------------------------
-
-        Ouput:
-
-        """
+    def compute_min_max(self, dashboard_cls):
         if type(dashboard_cls._cuxfilter_df.data) == dask_cudf.core.DataFrame:
```
bryevdv, line 85: This pattern of conditioning on `== dask_cudf.core.DataFrame` to compute min/max etc happens at least three times. Suggest factoring things to a common set of functions that hides the switching on the underlying type.

bryevdv (changes requested):

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -16,10 +107,273 @@ def _compute_array_all_bins(
...
+
+        if self.stride is None and self.data_points is not None:
+            if self.stride_type == int:
+                self.stride = self.stride_type(
+                    round((self.max_value - self.min_value) / self.data_points)
+                )
+            else:
+                self.stride = self.stride_type(
+                    (self.max_value - self.min_value) / self.data_points
+                )
```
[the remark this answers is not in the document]
bryevdv: or even:
```python
            raw_stride = (self.max_value - self.min_value) / self.data_points
            stride = round(raw_stride) if self.stride_type == int else raw_stride
            self.stride = self.stride_type(stride)
```
Just to be clear I am *not* advocating purely for shorter / more compressed code just for the sake of being shorter, but for isolating individual steps and doing them once, instead of spreading over multiple codepaths. In this case that is:

1. raw stride value
1. round (or not)
1. set from stride type

python/cuxfilter/assets/datetime.py
```diff
@@ -0,0 +1,123 @@
...
+    if _type in CUDF_DATETIME_TYPES and type(dates) in [list, tuple]:
+        dates = to_datetime(dates)
+        return type(dates)([date.to_datetime64() for date in dates])
+    return dates
+
+
+def check_series_for_nan(ser_):
+    if ser_.dtype == "float64":
+        return ser_[~cp.isnan(ser_)].shape[0] > 0
+    return True
```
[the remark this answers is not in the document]
AjayThorve: returns True if length of series without NaN rows is greater than 0, should actually be `if series.dtype in [int, float]:`.

python/cuxfilter/charts/panel_widgets/panel_widgets.py
```diff
@@ -52,8 +54,49 @@ def range_slider(
...
+    ----------
+
+    x: str
+        column name from gpu dataframe
+
+    width: int,  default 400
+
+    height: int,  default 20
+
+    data_points: int,  default None
```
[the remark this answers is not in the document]
AjayThorve: None signifies no custom number of bins provided and data_points will default to `df[self.x].nunique()`

python/cuxfilter/charts/core/core_chart.py
```diff
@@ -8,6 +8,8 @@ class BaseChart:
     chart_type: str = None
     x: str = None
     y: str = None
+    x_dtype: str = None
+    y_dtype: str = None
```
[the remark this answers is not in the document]
AjayThorve: converted x_dtype and y_dtype to read-only properties based off other comments

python/cuxfilter/charts/core/core_widget.py
```diff
@@ -21,6 +21,7 @@ class BaseWidget:
     params = None
     min_value: float = 0.0
     max_value: float = 0.0
+    x_dtype = None
```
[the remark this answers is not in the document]
AjayThorve: converted x_dtype to read-only property based off other comments

python/cuxfilter/assets/datetime.py
```diff
@@ -0,0 +1,123 @@
...
+        _type = dtype
+    -------------------------------------------
+
+    Ouput:
+        list of np.datetime64 objects
+
+    """
+    if _type in CUDF_DATETIME_TYPES and type(dates) in [list, tuple]:
+        dates = to_datetime(dates)
+        return type(dates)([date.to_datetime64() for date in dates])
```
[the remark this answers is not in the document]
AjayThorve: fixed

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -16,10 +107,273 @@ def _compute_array_all_bins(
...
+        if type(dashboard_cls._cuxfilter_df.data) == dask_cudf.core.DataFrame:
+            self.min_value = (
+                dashboard_cls._cuxfilter_df.data[self.x].min().compute()
+            )
+            self.max_value = (
+                dashboard_cls._cuxfilter_df.data[self.x].max().compute()
+            )
+        else:
+            self.min_value = dashboard_cls._cuxfilter_df.data[self.x].min()
+            self.max_value = dashboard_cls._cuxfilter_df.data[self.x].max()
```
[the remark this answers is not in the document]
AjayThorve: refactored and created a common _get_min_max() function in core_chart, which also checks for dask_cudf instance

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -16,10 +107,273 @@ def _compute_array_all_bins(
...
-        # -1 for 0-based indexing
-        np.put(result_array, update_data_x.astype(int) - 1, update_data_y)
-        return result_array[source_x.astype(int) - 1]
+        result_array = np.zeros(shape=source_x.shape)
+        indices = [np.where(x_ == source_x)[0][0] for x_ in update_data_x]
+        np.put(result_array, indices, update_data_y)
+        return result_array
+
+    def compute_min_max(self, dashboard_cls):
+        if type(dashboard_cls._cuxfilter_df.data) == dask_cudf.core.DataFrame:
```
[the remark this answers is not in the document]
AjayThorve: not as such, reading through some articles, `isinstance` seems like a better fit. replaced all instances of `type()` with `isinstance` wherever it made sense

python/cuxfilter/charts/core/aggregate/core_aggregate.py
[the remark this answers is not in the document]
AjayThorve: makes sense, thanks for the suggestion. Implemented it.

python/cuxfilter/charts/core/non_aggregate/core_line.py
```diff
@@ -95,21 +93,56 @@ def initiate_chart(self, dashboard_cls):
...
+
+        if self.data_points > len(dashboard_cls._cuxfilter_df.data):
+            self.data_points = len(dashboard_cls._cuxfilter_df.data)
+
+        if self.x_dtype == "bool":
+            self.min_value = 0
+            self.max_value = 1
+            self.stride = 1
+            # set axis labels:
+            dict_map = {0: "False", 1: "True"}
```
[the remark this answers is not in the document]
AjayThorve: added `BOOL_MAP` to `constants.py`

python/cuxfilter/charts/panel_widgets/panel_widgets.py
```diff
@@ -52,8 +54,49 @@ def range_slider(
...
+
+    step_size: np.timedelta64, defautl np.timedelta64(days=1)
```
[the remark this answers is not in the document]
AjayThorve: fixed

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -1,11 +1,102 @@
...
+        self.x = x
+        self.y = y
+        self.input_stride = step_size
+        self.stride_type = step_size_type
+        self.data_points = data_points
+        self.add_interaction = add_interaction
+        self.aggregate_fn = aggregate_fn
+        self.height = height
+        self.width = width
+        if len(title) == 0:
```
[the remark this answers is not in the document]
AjayThorve: fixed

python/cuxfilter/charts/core/non_aggregate/core_line.py
AjayThorve: fixed

python/cuxfilter/assets/datetime.py
AjayThorve: fixed

python/cuxfilter/assets/datetime.py
AjayThorve: converted self.x_dtype and self.y_dtype to read-only compyted properties  as seen [here](https://github.com/AjayThorve/cuxfilter/blob/new_dtypes/python/cuxfilter/charts/core/core_chart.py#L76-L90).

Also, transfered execution of all dt functions to BaseChart, so that something like `self.to_xaxis_type(event.geometry["x"])` works, all functions [here](https://github.com/AjayThorve/cuxfilter/blob/new_dtypes/python/cuxfilter/charts/core/core_chart.py#L132-L204)

python/cuxfilter/charts/core/non_aggregate/core_line.py
AjayThorve: fixed, added a common [method](https://github.com/AjayThorve/cuxfilter/blob/new_dtypes/python/cuxfilter/charts/core/core_chart.py#L206-L209) to BaseChart class

python/cuxfilter/charts/core/aggregate/core_aggregate_choropleth.py
```diff
@@ -255,10 +259,14 @@ def compute_query_dict(self, query_str_dict):
         if len(list_of_indices) == 0 or list_of_indices == [""]:
             query_str_dict.pop(self.name, None)
         elif len(list_of_indices) == 1:
-            query_str_dict[self.name] = self.x + "==" + str(list_of_indices[0])
+            query_str_dict[self.name] = "{}=={}".format(
+                self.x, list_of_indices[0]
+            )
```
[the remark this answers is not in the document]
AjayThorve: added f-strings throughout

python/cuxfilter/charts/core/aggregate/core_aggregate_choropleth.py
```diff
@@ -166,13 +166,17 @@ def _compute_array_all_bins(
...
+        indices = [np.where(x_ == source_x)[0][0] for x_ in update_data_x]
+        np.put(result_array, indices, update_data_y)
+        return result_array
+
+        # result_array = np.zeros(
+        #     shape=(int(max(source_x.max(), update_data_x.max())),)
+        # )
+        # # -1 for 0-based indexing, making sure indexes are type int
+        # np.put(result_array, (update_data_x - 1).astype(int), update_data_y)
+        # return result_array[source_x.astype(int) - 1]
```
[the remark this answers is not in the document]
AjayThorve: removed

exactlyallan (approved):

python/cuxfilter/charts/panel_widgets/panel_widgets.py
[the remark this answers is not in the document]
bryevdv: That seems like worth putting in the docstring here, unless there is some other place users can easily find that information?

python/cuxfilter/charts/datashader/datashader.py
```diff
@@ -104,6 +104,8 @@ def scatter(
     A cudashader scatter plot.
     Type cuxfilter.charts.datashader.custom_extensions.InteractiveImage
     """
+    if isinstance(color_palette, tuple):
+        color_palette = list(color_palette)
```
bryevdv, line 108: Can `Scatter`, `Graph` etc be made indifferent to the type of sequence a palette is stored in?

python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -130,14 +122,11 @@ def compute_stride(self):
...
-                    round((self.max_value - self.min_value) / self.data_points)
-                )
-            else:
-                self.stride = self.stride_type(
-                    (self.max_value - self.min_value) / self.data_points
-                )
+            raw_stride = (self.max_value - self.min_value) / self.data_points
+            stride = (
+                round(raw_stride) if self.stride_type == int else raw_stride
+            )
```
bryevdv, line 126: We have such a short mandated line length (no action suggested, just :cry:)

python/cuxfilter/charts/core/core_chart.py
```diff
@@ -203,6 +203,11 @@ def _yaxis_stride_type_transform(self, stride_type):
         """
         return dt.transform_stride_type(stride_type, self.y_dtype)
 
+    def _get_min_max(self, df, col_name):
+        if isinstance(df, dask_cudf.core.DataFrame):
+            return (df[col_name].min().compute(), df[col_name].max().compute())
+        return (df[col_name].min(), df[col_name].max())
```
bryevdv, line 209: Suggest avoid dupe code inside the method:
```
min, max = df[col_name].min(), df[col_name].max()

if isinstance(df, dask_cudf.core.DataFrame):
    return (min.compute(), max.compute())

return (min, max)
```

But also suggest just making this a free util function, not a method (then `core_widget.py`) won't need an identical method)

python/cuxfilter/charts/panel_widgets/plots.py
```diff
@@ -257,20 +237,12 @@ def initiate_chart(self, dashboard_cls):
...
-        else:
-            self.min_value = int(
-                dashboard_cls._cuxfilter_df.data[self.x].min()
-            )
-            self.max_value = int(
-                dashboard_cls._cuxfilter_df.data[self.x].max()
+        self.min_value, self.max_value = (
+            int(i)
+            for i in self._get_min_max(
+                dashboard_cls._cuxfilter_df.data, self.x
```
bryevdv, line 243: I usually like black formatting but this one took a second decipher with the short line length.  FWIW I would suggest `x` as a variable name for a value that is not an integer (and there are not integers *to start*). 

But maybe this would look clearer under black formatting?
```
min, max = self._get_min_max(dashboard_cls._cuxfilter_df.data, self.x)
self.min_value = int(min)
self.max_value = int(max)
```

python/cuxfilter/assets/datetime.py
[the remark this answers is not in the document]
bryevdv: `int` arrays can't have any NaN values ever, so that case would always return False unless I am misunderstanding

Edit: unless something is different for cupy?

python/cuxfilter/charts/core/core_chart.py
```diff
@@ -70,6 +73,22 @@ def height(self, value):
     def library_specific_params(self):
         return self._library_specific_params
 
+    @property
+    def x_dtype(self):
+        if isinstance(self.source, ColumnDataSource):
+            return self.source.data[self.data_x_axis].dtype
+        elif isinstance(self.source, (cudf.DataFrame, dask_cudf.DataFrame)):
+            return self.source[self.x].dtype
+        return None
```
bryevdv, line 82: No action here but comment: it seems like the best place to improve this is upstream, i.e. see if there is away for  CDS columns to be cudf series, or cupy arrays. It's possible a new (sub?) type to cope with deferred evaluation or device/host movement would be needed. Also the lib dependencies could only be optional, used if passed in but not otherwise imported. Maybe that's more trouble than it's worth but if you think it's worth exploring I'd say open a Bokeh issue to discuss.

python/cuxfilter/charts/core/core_chart.py
```diff
@@ -110,6 +129,80 @@ def _repr_mimebundle_(self, include=None, exclude=None):
...
+        return dt.to_dt_if_datetime(dates, self.x_dtype)
+
+    def _yaxis_dt_transform(self, dates):
+        """
+        Description: convert to datetime64 if self.y_dtype is of type datetime
+        -----------------------------------------------------------------
+        Input:
+            dates: list | tuple of integer timestamps objects
+        """
+        # self.x_dtype is a computed read-only property
```
bryevdv, line 169: should be `y_dtype` in the comment but I don't think these comments are especially necessary either.

bryevdv (changes requested):
few last comments

python/cuxfilter/assets/datetime.py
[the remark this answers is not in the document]
AjayThorve: ah yes, you are right,  my bad. So the condition will be `return True if length of series without NaN rows is greater than 0, when dtype == float`, fixing it

python/cuxfilter/charts/core/aggregate/core_aggregate.py
AjayThorve: I know :(

python/cuxfilter/charts/panel_widgets/plots.py
AjayThorve: fixed

python/cuxfilter/charts/core/core_chart.py
AjayThorve: Have made a note of it, will try exploring a bit into this, and create a separate issue on bokeh

python/cuxfilter/charts/core/core_chart.py
AjayThorve: fixed

python/cuxfilter/charts/core/core_chart.py
AjayThorve: fixed, added it as a free util function under assets.cudf_utils

python/cuxfilter/charts/panel_widgets/panel_widgets.py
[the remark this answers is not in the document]
AjayThorve: added it to the docstring

python/cuxfilter/assets/datetime.py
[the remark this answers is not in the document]
AjayThorve: fixed

python/cuxfilter/charts/datashader/datashader.py
AjayThorve: it's actually datashader that throws an error  [example](https://github.com/AjayThorve/cuxfilter/blob/new_dtypes/python/cuxfilter/charts/datashader/plots.py#L223) if the input color_palette is not a list, so we need to make sure it is a list during chart init.

AjayThorve: image

python/cuxfilter/assets/datetime.py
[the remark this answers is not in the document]
bryevdv: Is it guaranteed that that there will only ever be 64 bit float dtypes?
```
In [1]: import cupy as cp                                                                                                                              

In [2]: a32 = cp.array([1,2,3], dtype=cp.float32)                                                                                                      

In [3]: a64 = cp.array([1,2,3], dtype=cp.float64)                                                                                                      

In [4]: a64.dtype == float                                                                                                                             
Out[4]: True

In [5]: a32.dtype == float                                                                                                                             
Out[5]: False
```
If this needs to work on 32bit arrays I think it needs to use `issubdtype` to be more more general:
```
In [12]: np.issubdtype(a32.dtype, np.floating)                                                                                                         
Out[12]: True

In [13]: np.issubdtype(a64.dtype, np.floating)                                                                                                         
Out[13]: True
```

Also maybe there is a better function name? "check series for nan" sounds exactly like "return true if there is a nan in the series", but it's almost the opposite of that? What about `check_no_nan_runs`? That says (to me) "return true if there is not a run of nans"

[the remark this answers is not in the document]
bryevdv: actually is this just checking that the array is not *all* nans?
```
In [14]: not all(np.isnan(a32))                                                                                                                        
Out[14]: True

In [15]: not all(np.isnan(a64))                                                                                                                        
Out[15]: True

In [16]: ai = cp.array([1,2,3], dtype=cp.int32)                                                                                                        

In [17]: not all(np.isnan(ai))                                                                                                                         
Out[17]: True

```
?

python/cuxfilter/charts/datashader/datashader.py
bryevdv: That's a bit surprising, since I thought datashader could handle Bokeh palettes directly, but I guess not. But `Scatter` and `Graph` could do this check and update internally then? I'm asking about insulating callers of `Scatter` and `Graph` from the burden of having to know to massage the palettes in to some specific different format

bryevdv (approved):

PR #214, opened 2020-11-09
Patch fixes

AjayThorve:
This short PR implements a couple of fixes:

1. For aggregate chart, moving the initialization of `custom_binning` parameter to object init state, to avoid it being set after re-running the chart.initialize twice.

2. Fixed the map_style parameter of deck gl charts, to correctly identify user-input, and cuxfilter.themes if user input is None (default is dark map_style)

### DIFF
--- a/python/cuxfilter/charts/core/aggregate/core_aggregate.py
+++ b/python/cuxfilter/charts/core/aggregate/core_aggregate.py
@@ -37,6 +37,12 @@ class BaseAggregateChart(BaseChart):
             else:
                 self.filter_widget.bar_color = DATATILE_INACTIVE_COLOR
 
+    @property
+    def custom_binning(self):
+        return (
+            self._stride is not None or self._data_points is not None
+        )
+
     def __init__(
         self,
         x,
@@ -82,11 +88,9 @@ class BaseAggregateChart(BaseChart):
         """
         self.x = x
         self.y = y
-        if step_size or data_points:
-            self.custom_binning = True
-        self.stride = step_size
+        self._stride = step_size
+        self._data_points = data_points
         self.stride_type = step_size_type
-        self.data_points = data_points
         self.add_interaction = add_interaction
         self.aggregate_fn = aggregate_fn
         self.height = height
@@ -142,6 +146,11 @@ class BaseAggregateChart(BaseChart):
 
         """
         self.source = dashboard_cls._cuxfilter_df.data
+        # reset data_point to input _data_points
+        self.data_points = self._data_points
+        # reset stride to input _stride
+        self.stride = self._stride
+
         if self.x_dtype == "bool":
             self.min_value = 0
             self.max_value = 1
--- a/python/cuxfilter/charts/core/aggregate/core_aggregate.py
+++ b/python/cuxfilter/charts/core/aggregate/core_aggregate.py
@@ -23,6 +23,8 @@ class BaseAggregateChart(BaseChart):
     use_data_tiles = True
     custom_binning = False
     datatile_active_color = DATATILE_ACTIVE_COLOR
+    stride = None
+    data_points = None
 
     @property
     def datatile_loaded_state(self):
--- a/python/cuxfilter/charts/core/aggregate/core_aggregate.py
+++ b/python/cuxfilter/charts/core/aggregate/core_aggregate.py
@@ -82,7 +82,9 @@ class BaseAggregateChart(BaseChart):
         """
         self.x = x
         self.y = y
-        self.input_stride = step_size
+        if step_size or data_points:
+            self.custom_binning = True
+        self.stride = step_size
         self.stride_type = step_size_type
         self.data_points = data_points
         self.add_interaction = add_interaction
@@ -161,14 +163,6 @@ class BaseAggregateChart(BaseChart):
                 self.compute_stride()
             else:
                 self.use_data_tiles = False
-
-        if self.input_stride is None:
-            # No stride for bins specified, in this we case,
-            # we compute cudf.Series.value_counts() for histogram
-            self.custom_binning = False
-        else:
-            self.stride = self.input_stride
-            self.custom_binning = True
 
         self.calculate_source(dashboard_cls._cuxfilter_df.data)
         self.generate_chart()
--- a/python/cuxfilter/charts/core/aggregate/core_aggregate.py
+++ b/python/cuxfilter/charts/core/aggregate/core_aggregate.py
@@ -41,9 +41,7 @@ class BaseAggregateChart(BaseChart):
 
     @property
     def custom_binning(self):
-        return (
-            self._stride is not None or self._data_points is not None
-        )
+        return self._stride is not None or self._data_points is not None
 
     def __init__(
         self,
--- a/python/cuxfilter/charts/core/core_chart.py
+++ b/python/cuxfilter/charts/core/core_chart.py
@@ -24,7 +24,7 @@ class BaseChart:
     data_points: int = 0
     filter_widget = None
     _library_specific_params: Dict[str, str] = {}
-    _stride = None
+    stride = None
     stride_type = int
     min_value: float = 0.0
     max_value: float = 0.0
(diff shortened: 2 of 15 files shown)


python/cuxfilter/charts/core/aggregate/core_aggregate.py
```diff
@@ -82,7 +82,9 @@ def __init__(
         """
         self.x = x
         self.y = y
-        self.input_stride = step_size
+        if step_size or data_points:
+            self.custom_binning = True
```
bryevdv, line 86: Why not a computed read-only property for `custom_binning` so that it does not have to be manually put in sync

AjayThorve: problem is `custom_binning` only needs to be set when chart is declared, and it should be static throughout. Computed would make sense when `custom_binning` is expected to change throughout the lifetime of the dashboard, right?

Now its only updated when chart object is created (cuxfilter.charts.bar/line). the previous logic updated it whenever dashboard was initialized.

AjayThorve: only if one of input `data_points` or `step_size` exist, will `custom_binning` be True. `sefl.data_points` and `self.stride` are updated when we initialize the dashbooard (chart.initiate_chart is called), and `custom_binning` can get out of sync if we run those cells twice.

python/cuxfilter/charts/deckgl/plots.py
```diff
@@ -242,3 +240,8 @@ def apply_theme(self, properties_dict):
                 "color_palette"
             ]
             self.compute_colors()
+        if self.map_style is None:
+            bg_color = properties_dict["map_style"]
+            self.chart._deck.map_style = (
+                f"mapbox://styles/mapbox/{bg_color}-v9"
```
bryevdv, line 246: There appears to be a `dark-v10` if this link is accurate:

https://docs.mapbox.com/mapbox-gl-js/example/setstyle/

what happens if a users wants to use that? In general I think hardcoding the bits like `-v9` out of reach of users has the potential to frustrate. I would say it should always be accepted to specify any full mapbox name eg. `dark-v9` or `dark-v10` and if you want to have convenience aliases so that e.g. `foo` really means `foo-v9` that is fine (but should be explicitly communicated) 

I'm also confused about the link to the gallery in the doctoring above. "dark" does not appear anywhere on that page. Id "Decimal" (on that page) a valid value to pass as a style? (Also: does case matter?)

python/cuxfilter/charts/core/aggregate/core_aggregate.py
bryevdv: >  and it should be static throughout. 

I would say that is another point in favor of a getter-only property -- it can't accidentally / unintentionally set

>  and custom_binning can get out of sync if we run those cells twice.

I guess I don't understand the full codepath. You are saying it is normal operation for there to be situations where `custom_binning` is True, but yet `data_points` and `stride` become un-set?

python/cuxfilter/charts/deckgl/plots.py
AjayThorve: actually that is just the theme default value, the user input for `map_style` is None, then map_style for the map is set as per the theme.


bg_color = "dark" for dark theme and rapids theme, and is "light" for light theme.


For the user input, we expect the full url as a string, like `mapbox://styles/mapbox/dark-v9`, or `mapbox://styles/mapbox/dark-v10`. Those urls work directly.

if None, pydeck defaults to its own default value, which as of version 0.5 is `mapbox://styles/mapbox/dark-v9`.

python/cuxfilter/charts/core/aggregate/core_aggregate.py
AjayThorve: okay, I get your point now, I somehow missed the `read-only` part,  `read-only getter` does make sense. I'll update it.

python/cuxfilter/charts/deckgl/plots.py
bryevdv: Where/how is `properties_dict["map_style"]` set? Are it's only valid values `"light"` and `"dark"`? If so then I think the docstring above could reflect that the default map style is actually dependent on light vs dark, i.e. 

* `properties_dict["map_style"] == "light"` --> default is `mapbox://styles/mapbox/light-v9`
* `properties_dict["map_style"] == "dark"` --> default is `mapbox://styles/mapbox/dark-v9`

Is that correct?

That said, does this mean that the `map_style` parameter to `choropleth` and the `"map_style"` value in `properties_dict` can't both accept the same set of values?

AjayThorve: actually `properties_dict` is not exposed to user when intializing the chart. Its actually imported when a theme is applied,and is mentioned here - https://github.com/AjayThorve/cuxfilter/blob/patch_fixes/python/cuxfilter/themes/dark.py#L152,

https://github.com/AjayThorve/cuxfilter/blob/patch_fixes/python/cuxfilter/themes/light.py#L152

AjayThorve: so when you call `cux_df.dashboard([charts], theme=cuxfilter.themes.dark)`, all charts are intiailized, `chart.initiate_chart` is called, and all themes are applied, so `chart.apply_theme` is called before starting the dashboard

AjayThorve: I agree with you, the themes file should have the entire URL, instead of just the keywords `dark` and `light`

bryevdv: In that case why not just put the real values `mapbox://styles/mapbox/dark-v9` and `mapbox://styles/mapbox/light-v9` in the dark/light them values so the f-string substitution can be removed (I just don't see the value of making that arbitrary transform in code if the theme can just have the correct value to begin)

AjayThorve: makes sense, I'll update the theme files with correct values

bryevdv: > so when you call cux_df.dashboard([charts], theme=cuxfilter.themes.dark), all charts are intiailized

Ok, I see. Is the right POV to look at this then that the "dark" is the default theme? If that's the case then I think what's itching me is that individual defaults are specified "in place" as well as in the theme, and then if the theme ever changes it's hard to keep those in sync. Maybe the docs could delegate e.g. "The default value is the "dark" theme map style" then if the theme move to `-v10` (or whatever) then the docstrings are still true.

python/cuxfilter/charts/core/aggregate/core_aggregate.py
AjayThorve: let me know if this looks good, `custom_binning` is now a read-only computed property, `self._stride` and `self._data_points` are internal variables, which store input stride and input data_points, based on which `self.custom_binning` is computed.

`self.stride` and `self.data_points` are updated based on the `calculate_source` fn, and are updated whenever chart source is calculated (used to calculate data_tiles and range_slider step size)

python/cuxfilter/charts/core/core_chart.py
```diff
@@ -24,7 +24,7 @@ class BaseChart:
     data_points: int = 0
     filter_widget = None
     _library_specific_params: Dict[str, str] = {}
-    _stride = None
+    stride = None
```
AjayThorve, line 27: Removed a property from core_charts `self._stride` getter and setter, which was redundant (previously had some logic which was later moved to `compute_stride` fn

python/cuxfilter/charts/deckgl/plots.py
AjayThorve: fixed this, the default value will actually be:

```python
"""
default based on cuxfilter.themes:
            dark/rapids theme: 'mapbox://styles/mapbox/dark-v9'
            light theme: 'mapbox://styles/mapbox/light-v9'
"""
```

bryevdv (approved):

PR #216, opened 2020-11-16
Patch fix for panel >0.10.0

AjayThorve:
Minor changes - added keyword arguments for get_server. `address` parameter was getting assigned as one of the positional arguments, when it should have been a default value, which was leading to a `gaierror` bug when running `cux_df.dashboard()` function.


GPUtester:

bryevdv (approved):

PR #227, opened 2021-01-26
Refactor themes

AjayThorve:
This PR focusses on refactoring the bloated `apply_theme` functions applied individually on all charts and transferring the logicx to css files + bokeh styling dictionaries applied just before running the dashboard.

Apart from the above, the PR also adds the following:
- Adds missing css and html template files to the package build

- Updated Datasize indicator (with themes):
image

- Fixed Loading Animation issue for Datashader+Bokeh charts by adding the selection_event callback to `chart.document.add_next_tick_callback` and manually setting `panel.state.busy`  = `True` while the callback is being executed.


python/cuxfilter/charts/core/non_aggregate/core_graph.py
```diff
@@ -115,7 +115,12 @@ def __init__(
         self.edge_aggregate_col = edge_aggregate_col
         self.node_aggregate_fn = node_aggregate_fn
         self.edge_aggregate_fn = edge_aggregate_fn
-        self.node_color_palette = list(node_color_palette)
+        if node_color_palette is None:
+            self.no_colors_set = True
+            self.node_color_palette = CUXF_DEFAULT_COLOR_PALETTE
+        else:
+            self.node_color_palette = list(node_color_palette)
```
bryevdv, line 122: Explicitly coordinating three values (parameter `node_color_palette`, the `no_colors_set` flag, and `self.node_color_palette`) in several places seems a bit fragile against future changes or additions. The amount of explicit state to manually set in the initializer could be reduced with some properties

```python
def __init__(self, ..., node_color_palette=None):
    self._node_color_palette_input = node_color_palette

@property
def colors_set(self):
    return self._node_color_palette_input is not None

@property 
def node_color_palette(self):
    if self.colors_set:
        return list(self._node_color_palette_input)
    return CUXF_DEFAULT_COLOR_PALETTE
```

Maybe not for now, but in the future it might be nice to partition the `node` and `edge` parts of this in to sub-objects. This palette handling code is almost-repeated below but can't be factored in to a mixin or base class easily because of the need to differentiate <pre>self.<b>node</b>_color_palette // self.<b>edge</b>_color_palette</pre> directly in the variable names. But things were split a little the palette naming could be generic via `self.node.color_palette` and `self.edge.color_palette`, and then all the `color_palette` handling could be factored out.

python/cuxfilter/themes/light.py
```diff
@@ -1,153 +1,90 @@
...
+from bokeh import palettes
+from panel.template.theme import DefaultTheme
+from ..layouts.layouts import ReactTemplate
+from ..charts.constants import DATATILE_ACTIVE_COLOR
 
-                .nav-container {
-                    margin-top: 1em;
-                    font-size: 1.04em;
-                }
+ASSETS_PATH = pathlib.Path(__file__).parent / "assets"
```
bryevdv, line 10: This is defined separately in a few modules now, maybe pull up to one place in `themes/__init__.py` ?

bryevdv (changes requested):

python/cuxfilter/themes/light.py
AjayThorve: Actually, @bryevdv  adding it to the `themes/__init__.py` is throwing a circular import error

python/cuxfilter/charts/core/non_aggregate/core_graph.py
AjayThorve: fixed this

python/cuxfilter/themes/light.py
AjayThorve: Added all static paths to `charts/constants.py` where the other constants reside

python/cuxfilter/charts/core/aggregate/core_datasize_indicator.py
```diff
@@ -43,9 +43,10 @@ def initiate_chart(self, dashboard_cls):
         self.generate_chart()
 
     def view(self):
+        import panel as pn
```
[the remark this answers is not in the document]
AjayThorve: fixed

python/cuxfilter/charts/core/aggregate/core_aggregate_choropleth.py
```diff
@@ -113,7 +113,7 @@ def __init__(
         self.tooltip = tooltip
         self.tooltip_include_cols = tooltip_include_cols
         self.nan_color = nan_color
-        self.title = title if title else f"choropleth for {self.x} "
+        self.title = title if title else f"{self.x}"
```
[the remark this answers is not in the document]
AjayThorve: fixed

bryevdv (approved):

PR #261, opened 2021-03-17
Bug fix

AjayThorve:
Minor fix - correcting the parameter name in core_choropleth from `datatile_dict` to `datatile`, which caused `dashboard.py#L757` to fail, as all other `query` functions have parameter `datatile`


bryevdv (approved):

AjayThorve:
@gpucibot merge

PR #303, opened 2021-07-01
Fix/update bokeh version

AjayThorve:
Fixes #254.

Follow up PR will remove support for custom datashader extension and support the much more exhaustive version that holoviews implements

PR #304, opened 2021-07-10
Fix/follow up to #303

AjayThorve:
Follow up PR to #303, fixes theme related bugs, which were introduced in the previous PR, while updating panel and bokeh version support. 

This PR also fixes(handles), a weird issue where the bokeh x-axis data in CDS is converted from datetime to float during runtime, which caused a bunch of issues in the computing queries.

Will have to investigate further and try and fix this issue in the base libraries instead of cuxfilter though, more work to follow regarding that.

### DIFF
--- a/python/cuxfilter/charts/core/aggregate/core_aggregate.py
+++ b/python/cuxfilter/charts/core/aggregate/core_aggregate.py
@@ -1,8 +1,6 @@
 import panel as pn
 import numpy as np
-from bokeh.models import DatetimeTickFormatter, ColumnDataSource
-import cudf
-import dask_cudf
+from bokeh.models import DatetimeTickFormatter
 
 from ..core_chart import BaseChart
 from ....assets.numba_kernels import calc_groupby, calc_value_counts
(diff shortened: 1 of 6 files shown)


ajschmidt8:
rerun tests

bryevdv (approved):

bryevdv:
> This PR also fixes(handles), a weird issue where the bokeh x-axis data in CDS is converted from datetime to float during runtime, which caused a bunch of issues in the computing queries

Definitely would like to understand whatever is going on with this. I have not seen any other reports about types mutating and I can't find any 2.3+ PRs that seem relevant either

AjayThorve:
> > This PR also fixes(handles), a weird issue where the bokeh x-axis data in CDS is converted from datetime to float during runtime, which caused a bunch of issues in the computing queries
> 
> Definitely would like to understand whatever is going on with this. I have not seen any other reports about types mutating and I can't find any 2.3+ PRs that seem relevant either

So its more of a panel issue, whenever we start a dashboard with panel, the bokeh charts with datetime axis gets converted to float. 

bokeh.io.show works fine, and does not have that issue

AjayThorve:
rerun tests

AjayThorve:
rerun tests

AjayThorve:
@gpucibot merge

PR #317, opened 2021-08-20
Fix/unique names

### LINKED ISSUE #316: [BUG] Same plot type not plotted in multi-plot dashboard
Plotting two plots of the same type is not working for displaying both plots in the dashboard, only the last plot will plot

Here's the base case:
image

Now when we change one of the plots to a bokeh line chart.. the names are different and the second plot will show.

image

Notice how the name of the charts is the same....
In `dashboard.py` https://github.com/rapidsai/cuxfilter/blob/branch-21.10/python/cuxfilter/dashboard.py#L190
```
        for chart in charts:
            self._charts[chart.name] = chart
            chart.initiate_chart(self)
            chart._initialized = True
```
So the same chart name is causing an overwrite in `self._charts[chart.name]`

AjayThorve:
fixes #316 

cc @esnvidia

### DIFF
--- a/python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
+++ b/python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
@@ -29,8 +29,10 @@ class BaseNonAggregate(BaseChart):
     def name(self):
         # overwrite BaseChart name function to allow unique chart on value x
         chart_type = self.chart_type if self.chart_type else "chart"
-        return f"""{self.x}_{self.y}{'_'+self.aggregate_col or ''}
-        {'_'+self.aggregate_fn or ''}_{chart_type}_{self.title}"""
+        return (
+            f"{self.x}_{self.y}_{self.aggregate_col or ''}_"
+            f"{self.aggregate_fn or ''}_{chart_type}_{self.title}"
+        )
 
     def initiate_chart(self, dashboard_cls):
         """
--- a/python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
+++ b/python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
@@ -30,8 +30,10 @@ class BaseNonAggregate(BaseChart):
         # overwrite BaseChart name function to allow unique chart on value x
         chart_type = self.chart_type if self.chart_type else "chart"
         return (
-            f"{self.x}_{self.y}_{self.aggregate_col or ''}_"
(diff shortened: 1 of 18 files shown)


python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
```diff
@@ -29,7 +29,10 @@ class BaseNonAggregate(BaseChart):
     def name(self):
         # overwrite BaseChart name function to allow unique chart on value x
         chart_type = self.chart_type if self.chart_type else "chart"
-        return f"{self.x}_{self.y}_{chart_type}"
+        return (
+            f"{self.x}_{self.y}_{self.aggregate_col or ''}_"
+            f"{self.aggregate_fn or ''}_{chart_type}_{self.title}"
```
bryevdv, line 34: Just noting that 
```
{self.aggregate_col or ''}_{self.aggregate_fn or ''}_
```
seems like it could end up leaving double underscore `__` in the name (maybe that's fine, just noting in case)

python/cuxfilter/dashboard.py
```diff
@@ -96,6 +97,19 @@ def _create_app(
     return server
 
 
+def _check_if_duplicates(charts):
+    _charts = [i.name for i in charts]
+    dups = list((Counter(_charts) - Counter(set(_charts))).keys())
+    if len(dups) > 0:
+        print(
```
bryevdv, line 104: `warning.warn`? 

Also I think `Counter` can be applied moe simply:
```

In [11]: from collections import Counter

In [12]: data = ["foo", "bar", "baz", "foo"]

In [13]: dups = [k for k, v in Counter(data).items() if v > 1]

In [14]: dups
Out[14]: ['foo']

bryevdv: Also `if dups` will suffice

AjayThorve: thanks, fixed this. Also using warnings.warn now

bryevdv (approved):

python/cuxfilter/charts/core/non_aggregate/core_non_aggregate.py
AjayThorve: yep, was aware of it. The aim is just to have unique names. 

On second thought, might as well be worth it to handle the extra `_` to avoid any future confusions

AjayThorve:
rerun tests

AjayThorve:
rerun tests

AjayThorve:
@gpucibot merge

PR #345, opened 2022-01-20
fix reinit function

AjayThorve:
Quick fix PR to fix the `_reinit_all_charts` function, which was no longer working due to the recent refactors

### DIFF
--- a/python/cuxfilter/dashboard.py
+++ b/python/cuxfilter/dashboard.py
@@ -319,10 +319,6 @@ class DashBoard:
         for chart in self.charts.values():
             chart.initiate_chart(self)
             chart._initialized = True
-            if chart.is_widget:
-                self._sidebar[chart.name] = chart
-            else:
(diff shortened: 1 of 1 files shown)


bryevdv:
@AjayThorve is there a good test case or example to run to see before/after? Changes look OK at a glance but I don't really have any context for what exactly was broken

AjayThorve:
@bryevdv (pushed one more change after you reviewed). basically don't need to reinitialize `data_size_indicator`, separately, as that was fixed in a previous PR. 

Current implementation was reinitializing it twice (line 320, 325), and this was leading to some inconsistencies in data size values if we ran `d.app()` twice, or `d.show()` twice in a row. 

Ideally, it should just reinitialize all chart if we rerun the function, which now happens properly

bryevdv (approved):

AjayThorve:
@gpucibot merge

AjayThorve:
rerun tests

PR #355, opened 2022-03-22
Update `bokeh` version

ajschmidt8:
This PR updates the `bokeh` version specification to be compatible with the specification that's used by `dask=2022.03.0` ([src](https://github.com/dask/dask/blob/2022.03.0/setup.py#L20)).


sevagh (approved):

ajschmidt8:
It seems like there's a bit of a circular dependency here, where the `rapids` package in the `integration` repo depends on `cuxfilter` (and these new changes), but `cuxfilter` requires the `bokeh` version spec in the `rapids-notebook-env` package from the `integration` repo to be updated. We have some upcoming changes to our pipelines that should address this problem in the future, but for now I'm going to admin merge these changes since everything else built correctly and @AjayThorve tested these changes locally.

PR #359, opened 2022-03-24
Fix/dask cudf support

AjayThorve:
This PR aims at fixing issues related to dask_cudf + cuxfilter compatibility, mainly:
- adds proper support for non-aggregate charts to use dask_cudf dataframes
- fixes dask_cudf issues with aggregate charts, including `view_dataframe`
- optimizes graph charts to support partition-based edge computations and native datashader support
- adds 80+ unit tests for existing core chart functions with dask_cudf

PR #366, opened 2022-04-14
fix dask_cudf issues and documentation

### LINKED ISSUE #367: Scatterplot change colour transparency for selected/unselected points
While exploring data in a scatterplot, it'd be useful to mark already explored points to a different colour to avoid duplicated exploration. Is there a way to do that? 

Furthermore, it'd be useful to make unselected points transparent with a low alpha instead of erasing them entirely while selecting a cluster of points (current default behaviour). Is that possible with cuxfilter?

AjayThorve:
This PR is a follow up to #359:
- fixes a couple of minor issues related to lasso_select in a dask.distributed environment
- Adds example notebooks demonstrating dask_cudf support
- Update documentation and Readme to reflect dask_cudf usage

New Features added:
- added `unseleced_alpha` parameter to all datashader charts, which when enabled displays unselected data as the same color palette but transparent (default=0.2)
- fixes #367.
Recording 2022-06-01 at 14 29 13

- added binary data transfer support for choropleth charts (using deckgl + bokeh.models.ColumnDataSource + panel's html+Deck widget), instead of the current method using pydeck, where huge json strings were sent to the client on every interaction, resulting in a visual lag whenever a fairly large amount of data points were rendered.


AjayThorve:
> Looks good at a very cursory glance. My only comment is that a param named `transparent_unselected_data` with type `bool` really boxes in future (and current) possibilities. `unselected_alpha` with type `float` would let the user decide the amount of alpha to apply and also leave open the door for other related params in the future, if they are needed, e.g. `unselected_color`, `unselected_spread`, etc.

That's a great suggestion @bryevdv , I updated the api to reflect the changes, `unselected_alpha=0` will mimic the behavior of no bg, and can be adjusted as needed (with 0.2 being default)

exactlyallan (approved):
LGTM - dask-cudf cuxfilter ftw.

AjayThorve:
@gpucibot merge

PR #368, opened 2022-04-21
Use conda to build python packages during GPU tests

Ethyling:
This PR convert the `from sources` build we are doing in GPU test job to a `conda build`. This is done for the following reasons:
- This is required step to improve the Ops CI/CD setup to a more convenient pipeline
- This is required to start using `conda` compilers and `mamba` to build RAPIDS packages
- This prevent us from manually managing and installing the dependencies in GPU job
- This ensure the packages can be installed
- This ensure the tests are running and working against the package content and not the build results. Currently the Python packages are not tested.

This may increase the global pipeline time, but the usage of `mamba` should resolve this as `mamba` is faster than `conda` to build packages

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# COPYRIGHT (c) 2020, NVIDIA CORPORATION.
+# COPYRIGHT (c) 2020-2002, NVIDIA CORPORATION.
 ##############################################
 # cuXfilter GPU build and test script for CI #
 ##############################################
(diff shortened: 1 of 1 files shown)


Ethyling:
rerun tests

Ethyling:
rerun tests

AjayThorve (approved):
looks good, thanks!

ci/gpu/build.sh
```diff
@@ -53,37 +55,27 @@ nvidia-smi
...
-# BUILD - Build cuxfilter from source
+# BUILD - Build cuxfilter
 ################################################################################
 
-gpuci_logger "Build cuxfilter"
-"$WORKSPACE/build.sh" clean cuxfilter
+gpuci_logger "Build and install cuxfilter"
+cd "${WORKSPACE}"
+gpuci_conda_retry build conda/recipes/cuxfilter --python=$PYTHON
+gpuci_mamba_retry install cuxfilter
```
ajschmidt8, line 75: We need to add a `--channel`/`-c` flag here to ensure that `conda` installs `cuxfilter` from the local filesytem.

ajschmidt8 (changes requested):

ci/gpu/build.sh
Ethyling: PR updated

ajschmidt8 (approved):

Ethyling:
@gpucibot merge

PR #402, opened 2022-09-06
fix test failing on non-matching indices for newer dask version

AjayThorve:
When upgrading to dask 22.9.0, this test was failing as a result of a dask_dataframe compute() call resulting in a cuDF dataframe with incorrect indices (keeps partitioned indices after `map_partitions` call). 

This does not affect the behavior of graph charts in cuxfilter, so a trivial fix to the test, which compares against a result dataframe with sequential indices, is enough to resolve the issue. 

cc @galipremsagar


galipremsagar (approved):
Thanks @AjayThorve !

exactlyallan (approved):

AjayThorve:
@gpucibot merge

PR #451, opened 2023-02-15
Update datashader version

AjayThorve:
Since datashader merged the changes fixing cupy array compatiblity with `datashader.transfer_functions.dynspread` and `spread`, it's safe to use the latest stable datashader version, and the inbuilt function, instead of a custom `dynspread` function which was handling `cupy` array edge case

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,23 +13,23 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath('...'))
+sys.path.insert(0, os.path.abspath("..."))
 
 
 # -- Project information -----------------------------------------------------
 
-project = 'cuxfilter'
-copyright = '2019, NVIDIA'
-author = 'NVIDIA'
+project = "cuxfilter"
+copyright = "2019, NVIDIA"
+author = "NVIDIA"
 
 # The version info for the project you're documenting, acts as replacement for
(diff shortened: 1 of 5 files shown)


docs/source/conf.py
```diff
@@ -13,31 +13,32 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath('...'))
+sys.path.insert(0, os.path.abspath("..."))
```
ajschmidt8, line 16:
```suggestion
```

Since this file is being touched in this PR, can we update it to reflect that changes in the commit below?

See the commit message for additional details.


- https://github.com/rapidsai/cuspatial/pull/868/commits/1ca27370930687575f37f83d634738db59913d07

The `sys.path`, `import`, and commented lines can all be removed.

We'll just want to add a copyright header back after removing the other commented lines.

ajschmidt8 (commented):
Just one quick request.

docs/source/conf.py
AjayThorve: thanks, I did not realize these were never used, good catch. Although, we do require os package to execute `os.environ.get`

AjayThorve: resolved it by removing block using `os` since it's not being used.

exactlyallan (approved):
7% improvement wow!

PR #469, opened 2023-04-21
Dropping Python 3.8

divyegala:
This PR drops Python 3.8 in favor of 3.9.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   python-build:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@branch-23.06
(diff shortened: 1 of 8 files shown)


ajschmidt8 (approved):
suprisingly this PR passed CI despite the its upstream RAPIDS dependencies not having published packages for `23.06` `3.9`. It looks like it's picking up old stable `3.9` packages from `22.12`.

Seems like we should probably adjust our nightly workflows to remove the stable Anaconda.org channels for nightly/PR workflows.

that'll be a discussion for another day.

divyegala:
@ajschmidt8 can we merge this in that case?

vyasr:
@exactlyallan @AjayThorve could we get a Python review here, and a merge when ready? Thank you!

AjayThorve (approved):
LGTM!

Issue #452, opened 2023-02-15
[FEA] 3rd party dependency GPU tests


AjayThorve:
Work on a way to automate GPU testing for 3rd party dependencies with GPU support, mainly, datashader, holoviews and hvplot. Current plan of action involves:

- [x] GHA workflow to install and test the GPU functionality of these libraries (#456)
 - [ ] ~~lambda function to create an endpoint, which will trigger this workflow~~
 - [ ] ~~expose it to anaconda devs, to trigger this workflow on or before project releases~~
 - [ ] ~~subscribe to workflows via slack/teams~~

cc @exactlyallan

exactlyallan:
@raydouglass + @ajschmidt8  Could you review this approach to see if using lambda functions violates any security policies? I.E. external party triggering a PR.

ajschmidt8:
> @raydouglass + @ajschmidt8 Could you review this approach to see if using lambda functions violates any security policies? I.E. external party triggering a PR.

We might want to brainstorm on a better way for this. In hindsight, I'm not a fan of the Lambda solution that I suggested.

It involves deploying some additional AWS infrastructure that we'd have to manage.

Is a weekly or every-other-nightly cron job suitable?

If I recall correctly, I think you guys mentioned that we don't necessarily need to run it every night.

As long as the jobs don't take an unusually long amount of time to run, I don't see any harm in just running it on a scheduled basis.

exactlyallan:
@ajschmidt8 Scheduled jobs could be an option - but would this still be a thing that can test a PR?

AjayThorve:
> > @ajschmidt8 Scheduled jobs could be an option - but would this still be a thing that can test a PR?
> 
> Do you mean test a PR from one of the third-party repositories?

yep

ajschmidt8:
> > Do you mean test a PR from one of the third-party repositories?
> 
> yep

Yes, we could support that.

We can configure the job in such a way that it will run on a recurring basis against the default branch _or_ it can be manually triggered and passed a particular PR that it should test.

If I remember correctly, @AjayThorve has already set up the new workflow to work like that.

exactlyallan:
Okay, as long as there is at least an option. Its a bit annoying to manually trigger it, but what are you going to do... the results of the run are visible through Slack bot but is there any other option?

exactlyallan:
**New Proposal:**

- [x] A. GHA done with #456 
- [ ] B. Set recurring runs to catch dependencies issues
[...]

PR #478, opened 2023-05-22
GHA - external dependency testing workflow: add a schedule to run once every week

AjayThorve:
Based on the discussion in #452, adding a weekly schedule to run the external dependency testing workflow.

cc @exactlyallan @ajschmidt8

### DIFF
--- a/.github/workflows/test-external.yaml
+++ b/.github/workflows/test-external.yaml
@@ -29,4 +29,4 @@ jobs:
       node_type: "gpu-v100-latest-1"
       arch: "amd64"
       container_image: "rapidsai/ci:latest"
-      run_script: "./ci/test_external.sh ${{ inputs.external-project }} ${{ inputs.pr_number}}"
+      run_script: "./ci/test_external.sh ${{ inputs.external-project || 'all' }} ${{ inputs.pr_number }}"
--- a/.github/workflows/test-external.yaml
+++ b/.github/workflows/test-external.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       external-project:
-        description: "one of [datashader, holoviews, all]"
+        description: "Project:"
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/test-external.yaml
ajschmidt8, line 1: the last line of this file references input values:

```
${{ inputs.external-project }}
```

The inputs object will be empty when this job is triggered via a cron event.

Therefore, you'll need to provide a default value when the inputs are empty.

Something like this:

```
${{ inputs.external-project || 'default_value' }}
```

The string value must be wrapped in single quotes, not double quotes.

The doc page below has additional information on expressions in GitHub Actions for reference:

- https://docs.github.com/en/actions/learn-github-actions/expressions

AjayThorve: updated it to have a default value of 'all', thanks for the helpful link. The second input parameter is not required for the cron job, and hence left it as is. Let me know if this works

AjayThorve:
test failures seem unrelated to the changes in this PR, investigating now.

ajschmidt8 (commented):
Last comment:

Should we update these RAPIDS versions to the latest stable? `23.04`?

- https://github.com/rapidsai/cuxfilter/blob/branch-23.06/ci/test_external.sh#L10

I assume we want them to test against stable versions and not nightlies, right?

If we were testing against nightlies, we could add an entry to https://github.com/rapidsai/cuxfilter/blob/branch-23.06/ci/release/update-version.sh to keep them up-to-date.

AjayThorve:
> If we were testing against nightlies, we could add an entry to https://github.com/rapidsai/cuxfilter/blob/branch-23.06/ci/release/update-version.sh to keep them up-to-date.

Great idea, thanks! Pushed the changes

ajschmidt8 (approved):

PR #479, opened 2023-05-30
fix tests failing due to unsorted results

AjayThorve:
This PR:
- fixes a couple of tests which were failing due to recent changes, leading to aggregate results being unsorted
- resolves docs build failures, which were caused by datashader rendering issues with latest numpy(1.24.3), fixed by adding a pin for numpy (<1.24, similar to the version pins in cugraph). @ajschmidt8, could you check if adding this version pin is fine?

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -23,6 +23,7 @@ dependencies:
 - nbsphinx
 - nodejs>=14
 - notebook>=0.5.0
+- numpy>=1.21,<1.24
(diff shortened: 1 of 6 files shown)


exactlyallan (approved):

jakirkham:
Asked OPS for a review offline

AjayThorve:
Numpy version 1.24.3 issue with datashader is fixed (https://github.com/holoviz/datashader/pull/1176), but not yet released. Tracking it here, so that the version pin for numpy can be reverted once datashader releases 0.14.5

raydouglass (approved):

PR #499, opened 2023-07-05
cuxfilter: Build CUDA 12 packages

### LINKED ISSUE #450: cuxfilter: CUDA 12 Conda Packages
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.04. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Toolkit dependencies from Conda
```

```[tasklist]
### Other dependencies from Conda using CUDA
- [ ] https://github.com/conda-forge/cupy-feedstock/issues/139
- [ ] https://github.com/conda-forge/numba-feedstock/issues/112
```

AjayThorve:
This PR builds cuxfilter conda packages using CUDA 12. Resolves #450.


raydouglass (approved):

conda/recipes/cuxfilter/meta.yaml
```diff
@@ -15,19 +15,20 @@ source:
...
     - bokeh >=2.4.2,<=2.5
-    - cudf ={{ minor_version }}
+    - cudf ={{ version }}
     - cupy >=12.0.0
-    - cuspatial ={{ minor_version }}
-    - dask-cuda ={{ minor_version }}
-    - dask-cudf ={{ minor_version }}
+    - cuspatial ={{ version }}
+    - dask-cuda ={{ version }}
+    - dask-cudf ={{ version }}
```
jakirkham, line 31: Wouldn't we want to keep `minor_version` here to allow for patch releases of these dependencies to be picked up?

jakirkham (commented):
Thanks Jay! 🙏

Generally this looks good. Had a question below

conda/recipes/cuxfilter/meta.yaml
AjayThorve: I picked it up from how dask-cudf had the recipe, but I get your point, if that's to accomadate for patch releases, then I agree, it's needed. I'll revert the changes. Thanks!

bdice (changes requested):

conda/recipes/cuxfilter/meta.yaml
AjayThorve: fixed it!

dependencies.yaml
```diff
@@ -88,7 +96,7 @@ dependencies:
           - notebook>=0.5.0
       - output_types: [conda]
         packages:
-          - cugraph=23.08
+          - cugraph=23.08.*
```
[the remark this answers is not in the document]
AjayThorve: updated it, thanks!

dependencies.yaml
```diff
@@ -88,7 +96,7 @@ dependencies:
           - notebook>=0.5.0
       - output_types: [conda]
         packages:
-          - cugraph=23.08
+          - cugraph==23.8.*
```
jakirkham, line 99: Should this be?

Also see other places we might want to make this change

```suggestion
          - cugraph==23.08.*
```

jakirkham (commented):
Thanks Ajay! 🙏

Had a question below

dependencies.yaml
bdice: No, we want PEP440 compatible pinnings here. https://github.com/rapidsai/cuxfilter/pull/499#discussion_r1258318646

We want to use normalized versions/pinnings that are compatible with both conda and pip. The current state `==23.8.*` is how we do that. See also [cudf's dependencies](https://github.com/rapidsai/cudf/blob/3c51c9ecb56c7130f300bcb21d4ca38a62cbd418/dependencies.yaml#L216-L217) and the format used by `update-version.sh` in various RAPIDS repos.

bdice: The places where we use "23.08" are places like `__version__`, `CMakeLists.txt`, `README.md`, Doxygen, etc. We have adopted the normalized convention for all conda/pip pinnings.

Compare the usage of `NEXT_SHORT_TAG_PEP440` to the use of `NEXT_SHORT_TAG` in cudf, for example: https://github.com/rapidsai/cudf/blob/branch-23.08/ci/release/update-version.sh

conda/environments/all_cuda-120_arch-x86_64.yaml
```diff
@@ -0,0 +1,45 @@
...
+channels:
+- rapidsai
+- rapidsai-nightly
+- conda-forge
+- nvidia
+dependencies:
+- bokeh>=2.4.2,<=2.5
+- cuda-version=12.0
+- cudf==23.8.*
+- cugraph==23.8.*
```
bdice, line 12: cugraph isn't ready for CUDA 12 yet. Should we hold off on merging this PR until the dependencies are ready? Or do we ignore that because it's only needed for notebooks? The environment file won't be solvable until that is ready, so we might need a workaround in `dependencies.yaml` (e.g. don't generate the CUDA 12 environment, or drop cugraph from CUDA 12 environments, etc.).

bdice (approved):
One question about cugraph notebook dependency (not yet available for CUDA 12), otherwise LGTM!

conda/environments/all_cuda-120_arch-x86_64.yaml
jakirkham: If it is only needed for notebooks, would suggest we go ahead and get this PR in. We can test notebooks later

dependencies.yaml
jakirkham: Ok sounds good. Wasn't obvious whether the change was intended

jakirkham (approved):

dependencies.yaml
AjayThorve: Hey @jakirkham, it was an intended change based on one of the suggestions by @bdice :)

conda/environments/all_cuda-120_arch-x86_64.yaml
AjayThorve: ok, sounds good

AjayThorve:
Thanks @bdice, @jakirkham  and @raydouglass for the review. Going ahead and merging it

jakirkham:
Thanks all! 🙏

PR #500, opened 2023-07-07
Switch Docs to PyData Theme

exactlyallan:
Updated cuxfilter docs to use the pydata theme and better follow newer docs conventions. Updated some copy. 
addresses #459

PR #419, opened 2022-12-06
Fix broken symlink

ajschmidt8:
This PR fixes a broken symlink for `cuxfilter`.

PR #502, opened 2023-07-13
Fix broken symlink

raydouglass:
Follow up from #500 where a notebook was moved but the symlink in `notebooks` was not updated.

Related: #419


AjayThorve (approved):
Lgtm!

raydouglass:
Failures are unrelated to the PR change and this PR is required to unblock https://github.com/rapidsai/docker/pull/545, so I will admin merge it.

PR #506, opened 2023-07-24
Switch to new wheels pipeline

divyegala:


ci/build_wheel.sh
```diff
@@ -0,0 +1,24 @@
...
+source rapids-configure-sccache
+source rapids-date-string
+
+# Use gha-tools rapids-pip-wheel-version to generate wheel version then
+# update the necessary files
+version_override="$(rapids-pip-wheel-version ${RAPIDS_DATE_STRING})"
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+bash ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
```
ajschmidt8, line 15:
```suggestion
ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
```

Can we remove the `bash` keyword here?

This will require that the called script has a proper shebang (`#!/bin/bash`) and is executable (`chmod +x <file>`).

divyegala: Done!

ci/build_wheel.sh
```diff
@@ -0,0 +1,24 @@
...
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
+echo "The package name and/or version was modified in the package source. The git diff is:"
+git diff
+
+cd python
+
+# Hardcode the output dir
```
vyasr, line 21: I've removed this comment in most places. It's outdated and no longer applicable to anything here.
```suggestion
```

ci/test_wheel.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eou pipefail
+
+mkdir -p ./dist
```
vyasr, line 6: I'm wondering if this command is superfluous. @ajschmidt8 [we already have an `mkdir -p` in this tool](https://github.com/rapidsai/gha-tools/blob/main/tools/_rapids-download-from-s3#L14), that's the same directory that `rapids-download-wheels-from-s3` is ultimately using, right? If so, we could also remove this `mkdir -p` from all the wheel test scripts (future cleanup for all the merged PRs).

vyasr: I'm testing this in the rmm/cudf/cuspatial PRs too.

vyasr (changes requested):
Couple of minor suggestions for improvement, then I think we'll be good.

ci/test_wheel.sh
divyegala: So we just remove this line?

vyasr (approved):
LGTM

ajschmidt8 (approved):

PR #523, opened 2023-08-10
Don't append CUDA suffix to python pkg name

raydouglass:
By convention, the python package name should not include the CUDA suffix.

https://github.com/rapidsai/cuxfilter/blob/branch-23.08/ci/release/apply_wheel_modifications.sh#L13 handles adding the suffix when needed.

This fixes using the wrong suffix in CUDA 11 builds such as https://github.com/rapidsai/cuxfilter/actions/runs/5814188510/job/15763351108#step:9:376


AjayThorve (approved):
Thanks for taking care of this !!

bdice (approved):

PR #547, opened 2023-10-11
updates to external tests

AjayThorve:
This PR resolves the external tests still failing: https://github.com/rapidsai/cuxfilter/actions/runs/6485671242/job/17612261430.

Basically separately installing direct dependencies and then installing indirect dependencies from dependencies.yaml was downgrading rapids versions to 23.08 separately. This PR updates the dependencies to be in a single place (external_dependencies.yaml), while also adding it to `ci/release/update-version.sh`

There were a couple of dask.array tests in datashader which were failing when using `pytest`  with `--numprocesses` flag, which is resolved here by separately running that test without the flag.

cc @exactlyallan

### DIFF
--- a/EXTERNAL_TESTS.md
+++ b/EXTERNAL_TESTS.md
@@ -1,0 +1,57 @@
+# External Tests Workflow
+
+This document provides an overview of the GitHub Actions workflow (`.github/workflows/test-external.yaml`) and associated script (`ci/test_external.sh`) for running external tests on specified Python libraries, such as Datashader and Holoviews.
+
+## Purpose
+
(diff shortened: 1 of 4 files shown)


raydouglass (approved):
Approving since the `external_dependencies.yaml` file already exists.

But, a future enhancement should be to move the external dependencies env file to be a file key in dependencies.yaml instead.

AjayThorve:
> Approving since the `external_dependencies.yaml` file already exists.
> 
> But, a future enhancement should be to move the external dependencies env file to be a file key in dependencies.yaml instead.

I'll do a follow up PR with this, thanks!

PR #562, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI


dependencies.yaml
```diff
@@ -71,34 +72,37 @@ dependencies:
...
           - matrix:
-              cuda: "12.0"
+              cuda: "11.2"
             packages:
-              - cuda-version=12.0
+              - cuda-version=11.2
+          - matrix:
+              cuda: "11.4"
+            packages:
+              - cuda-version=11.4
```
bdice, line 86: It looks like we lost CUDA 11.5 here?
```suggestion
          - matrix:
              cuda: "11.4"
            packages:
              - cuda-version=11.4
          - matrix:
              cuda: "11.5"
            packages:
              - cuda-version=11.5
```

raydouglass (approved):

jameslamb:
1 build, `conda-python-tests / tests (11.4.3, centos7, amd64, 3.9, v100, earliest)`, is failing with stuff like this

```text
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/opt/conda/envs/test/lib/python3.9/site-packages/xdist/workermanage.py", line 386, in process_from_remote
INTERNALERROR>     warning_message = unserialize_warning_message(
...
INTERNALERROR>   File "/opt/conda/envs/test/lib/python3.9/asyncio/events.py", line 642, in get_event_loop
INTERNALERROR>     raise RuntimeError('There is no current event loop in thread %r.'
INTERNALERROR> RuntimeError: There is no current event loop in thread 'Dummy-8'.
```

([build link](https://github.com/rapidsai/cuxfilter/actions/runs/7490228408/job/20388851853?pr=562))

Not sure if it's relevant, but that is the only job in the matrix testing with an older driver.

I tried adding the CTK back to the environment for that job: https://github.com/rapidsai/cuxfilter/pull/562/commits/9ce9fc8094e14cc8d09b05436f39ad94470caa80.

Let's see if it helps.

AjayThorve (approved):

jameslamb:
From discussion offline with @AjayThorve ... I just pushed https://github.com/rapidsai/cuxfilter/pull/562/commits/84ce513bd76f0dfe68a673363b53f73d1b20d0e2 putting a ceiling on `panel`.

I did this in the bluntest way possible (applying it to all conda envs and `pyproject.toml`) just to test if it fixes the problem, but that might not be ideal. I'm especially skeptical of the `pyproject.toml` change... if that makes it out into wheels that users install, having a ceiling on `panel` is going to be disruptive.

If this works, I'm planning to remove that ceiling from `pyproject.toml`, do you agree @AjayThorve ?

AjayThorve:
> If this works, I'm planning to remove that ceiling from `pyproject.toml`, do you agree @AjayThorve ?

Yeah if this works, there haven't been any issues with other envs, and seems like the issue is running parallet pytests with updated panel is causing it. In that case, all we need to do is remove the current upper pin for panel from run dependencies and add a specific panel dependency with `>=1.0,<1.3.4` to `test_python` section (under just conda)

PR #569, opened 2024-02-08
[Ready for Review] Pandas 2.0 compatibility

### LINKED ISSUE #512: Check compatability with cuDF pandas 2.0 migration
Per https://github.com/rapidsai/cudf/pull/13535 check cuxfilter for issues with Pandas 2.0.

AjayThorve:
fixes #512

### DIFF
--- a/python/cuxfilter/charts/deckgl/bindings/panel_deck.py
+++ b/python/cuxfilter/charts/deckgl/bindings/panel_deck.py
@@ -45,7 +45,7 @@ class PanelDeck(param.Parameterized):
 
     @property
     def valid_indices(self):
-        return self.indices.intersection(self.data.index)
+        return list(self.indices.intersection(self.data.index))
 
     def get_tooltip_html(self):
         """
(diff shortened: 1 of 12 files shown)


jakirkham:
Thanks Ajay! 🙏

Could you please add "Fixes #512" to the OP?

python/cuxfilter/charts/deckgl/bindings/panel_deck.py
```diff
@@ -45,7 +45,7 @@ class PanelDeck(param.Parameterized):
 
     @property
     def valid_indices(self):
-        return self.indices.intersection(self.data.index)
+        return list(self.indices.intersection(self.data.index))
```
galipremsagar, line 48: I hope this conversion from device to host is an informed decision.

galipremsagar (approved):
Changes LGTM, one minor comment.

AyodeAwe (approved):

python/cuxfilter/charts/deckgl/bindings/panel_deck.py
AjayThorve: Ah I can see why that's confusing, but `self.data` here is actually a pandas dataframe here, which is converted from a cudf Dataframe post aggregations, for direct rendering, so technically there is no device to host conversion.

Can see the ref on L34 same file, where data is of type param.DataFrame(), which only accepts pandas as of now

galipremsagar: ah, thanks for the clarification, Ajay!

exactlyallan (approved):

jakirkham:
Thanks Ajay and Prem! 🙏

PR #578, opened 2024-03-06
Use public cudf APIs where possible

mroeschke:
cudf's `core` namespace is considered "private", so changing some usages that have public alternative.


AjayThorve:
Might have to fix the style checks, but overall looks good, thanks for this!

AjayThorve (approved):
Lgtm, pending style check fixes

PR #575, opened 2024-02-29
Generalize GHA selectors for pure Python testing

jakirkham:
To eliminate hard-coding, generalize the GHA workflow logic to select one build for testing. This should simplify future cuxfilter updates.

xref: https://github.com/rapidsai/build-planning/issues/25

PR #597, opened 2024-05-22
fix matrix_filter to include cuda_11.x

AjayThorve:
This PR fixes an issue caused by #575, which resulted in cu-11 wheels not being generated for cuxfilter.


bdice (changes requested):

bdice:
This can probably by closed now that #590 is merged. cc: @AjayThorve

PR #600, opened 2024-06-05
use rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Proposes introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts.


build.sh
```diff
@@ -111,6 +111,7 @@ if (( ${NUMARGS} == 0 )) || hasArg cuxfilter; then
     echo "8"
     if [[ ${INSTALL_TARGET} != "" ]]; then
         python setup.py build_ext --inplace
+        RAPIDS_DISABLE_CUDA=true \
         python setup.py install --single-version-externally-managed --record=record.txt
```
bdice, line 115: I think we usually indent the line following an explicit break like this.
```suggestion
        RAPIDS_DISABLE_CUDA=true \
            python setup.py install --single-version-externally-managed --record=record.txt
```

bdice (approved):
LGTM, just one minor comment.

build.sh
jameslamb: ok sure

AjayThorve (approved):

PR #607, opened 2024-07-11
Build and test with CUDA 12.5.1

KyleFromNVIDIA:
This PR updates the latest CUDA build/test version 12.2.2 to 12.5.1.

Contributes to https://github.com/rapidsai/build-planning/issues/73

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   python-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.08
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@cuda-12.5.1
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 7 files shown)


jameslamb (changes requested):
I think this should be updated to 12.5:

https://github.com/rapidsai/cuxfilter/blob/14237a1a911fd0ef48555db63c8f4b7b4f533cdc/docs/source/user_guide/installation.rst#L10-L12

(context: https://github.com/rapidsai/dask-cuda/pull/1357#pullrequestreview-2180844366)

----

This one is wrong (and looks like it has been since before this PR):

https://github.com/rapidsai/cuxfilter/blob/14237a1a911fd0ef48555db63c8f4b7b4f533cdc/docs/source/user_guide/installation.rst#L35-L42

It's has a comment that says "for CUDA 11.8", but then pulls a CUDA 12.0 image, and then the `docker run` after that `docker pull` is for an 11.8 image.

The `rapidsai/rapidsai` image isn't updated any more. Context: https://github.com/rapidsai/deployment/issues/382

Since that doc is encouraging looking at a notebook at `localhost:8888`, it should probably be using this image in both commands

```text
rapidsai/notebooks:24.06-cuda12.0-py3.10
```

(ref: https://hub.docker.com/r/rapidsai/notebooks/tags)

I just posted a separate question (does not need to block this PR) about updating those images at https://github.com/rapidsai/build-planning/issues/73.

jameslamb (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #620, opened 2024-08-14
Fix tests

AjayThorve:
This PR fixes

1. [failing tests](https://github.com/rapidsai/cuxfilter/actions/runs/10303789190/job/28671747182?pr=617), which were caused by duplicate columns in the core_graph source_df. 

2. Failing [external tests](https://github.com/rapidsai/cuxfilter/actions/workflows/test-external.yaml)


mroeschke (approved):
The Python change LGTM

python/cuxfilter/charts/core/non_aggregate/core_graph.py
```diff
@@ -50,7 +50,7 @@ def edge_columns(self):
 
     @property
     def node_columns(self):
```
bdice, line 52: I would rewrite this to avoid writing the first three column names twice.
```python
def node_columns(self):
    cols = [
        self.node_id,
        self.node_x,
        self.node_y,
    ]
    if self.node_aggregate_col != self.node_id:
        cols.append(self.node_aggregate_col)
    return cols
```

bdice (approved):
Looks fine, with one suggested refactor.

python/cuxfilter/charts/core/non_aggregate/core_graph.py
AjayThorve: Yep, agreed, this is better, applied

exactlyallan (approved):

PR #622, opened 2024-08-19
Remove NumPy <2 pin, remove panel ceiling

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).

EDIT: Sorry, converted back to draft again, NumPy 2 is not yet picked up.


msarahan (approved):

jakirkham:
Note that `cuxfilter` depends on `cuspatial`

https://github.com/rapidsai/cuxfilter/blob/ca63e9d5fdfced7f993f18989ae42409185cff43/dependencies.yaml#L200

So this depends on PR: https://github.com/rapidsai/cuspatial/pull/1441

jakirkham:
cc @jameslamb (for awareness)

seberg:
Yeah, sorry, I unmarked draft too early, but I think once things are in place and NumPy is picked up we can make it no-draft again and it should be good.

jakirkham:
All good. Was more looping in James as we had discussed the cuSpatial NumPy 2 PR earlier

jameslamb:
I've taken over https://github.com/rapidsai/cuspatial/pull/1441 and so am testing here too.

Got `cuspatial` to a state where it can be installed alongside Numpy (2.0.2): https://github.com/rapidsai/cuspatial/pull/1441#issuecomment-2359913946

I pushed commits here to use the packages built from that PR's CI. Let's see if it'll be enough to loosen the pin here!

jameslamb:
Putting an explicit `>=2.0` floor here, I see all builds and tests passing!

```text
  - numpy                                     1.26.4  py312heda63a1_0                      conda-forge             Cached
  + numpy                                      2.0.2  py312h58c1407_0                      conda-forge                8MB
```

([build link](https://github.com/rapidsai/cuxfilter/actions/runs/11003722592/job/30553411386?pr=622))

That's evidence that `cuxfilter` and its dependencies can be installed alongside NumPy 2.x.

Now that https://github.com/rapidsai/cuspatial/pull/1441 is merged, I just pushed commits removing all the testing-only changes here and moving that pin back down to `>=1.23,<3.0.0a0`.

Hopefully we'll see everything pass here and use NumPy 2.x.

dependencies.yaml
```diff
@@ -260,9 +260,6 @@ dependencies:
           - pytest
           - pytest-cov
           - pytest-xdist
-      - output_types: conda
-        packages:
-          - panel>=1.0,<=1.3.4
```
jakirkham, line 265: Interested to see how this goes

If we need to ignore a specific version, maybe we can do `!=`

Though if this is just the case of working around a past solver issue or limiting to a set of tested releases, then simply dropping seems reasonable

Maybe we can follow up with the cuxfilter team on which makes more sense

AjayThorve: Yeah, AFAIK, there are no issues at runtime, so as long as the tests pass (seems like they do), this should be fine

AjayThorve (approved):
LGTM!

jameslamb (approved):

jakirkham:
Thanks all! 🙏

PR #494, opened 2023-06-20
Refactor to use holoviews powered bar charts

AjayThorve:
This PR refactors the bokeh charts implementation, and uses holoviews to generate the charts, while also using box-selects instead of range_sliders. This was done for 2 reasons, stability of holoviews api, which also is much more efficient, and to use streams + custom callbacks to link all charts with bar_charts.

Some of the other notable changes:
- Using holoviews to generate bokeh plots, allows us to `str` columns directly for bar charts
- Persistent non-selection view for bar_charts
- Got rid of redundant bokeh line chart, and now a single implementation exists under datashader/line, callable `cuxfilter.charts.line`
- removed preview as pyppeteer was unstable
- Refactor to remove redundant functions and classes

PR #625, opened 2024-09-04
Add support for Python 3.12, drop pyppeteer dependency

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/40

This PR adds support for Python 3.12.

## Notes for Reviewers

This is part of ongoing work to add Python 3.12 support across RAPIDS.
It temporarily introduces a build/test matrix including Python 3.12, from https://github.com/rapidsai/shared-workflows/pull/213.

A follow-up PR will revert back to pointing at the `branch-24.10` branch of `shared-workflows` once all
RAPIDS repos have added Python 3.12 support.

Additional changes needed to add this support:

* dropping dependency on `pyppeteer` (https://github.com/rapidsai/cuxfilter/pull/625/files#r1750741079)

### This will fail until all dependencies have been updates to Python 3.12

CI here is expected to fail until all of this project's upstream dependencies support Python 3.12.

This can be merged whenever all CI jobs are passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,11 +158,11 @@ For nightly version `cuxfilter version == 24.10` :
 ```bash
 # for CUDA 12.0
 conda install -c rapidsai-nightly -c conda-forge -c nvidia \
-    cuxfilter=24.10 python=3.11 cuda-version=12.0
+    cuxfilter=24.10 python=3.12 cuda-version=12.0
 
 # for CUDA 11.8
 conda install -c rapidsai-nightly -c conda-forge -c nvidia \
-    cuxfilter=24.10 python=3.11 cuda-version=11.8
+    cuxfilter=24.10 python=3.12 cuda-version=11.8
 ```
 
 For the stable version of `cuxfilter` :
@@ -170,14 +170,14 @@ For the stable version of `cuxfilter` :
 ```bash
 # for CUDA 12.0
 conda install -c rapidsai -c conda-forge -c nvidia \
-    cuxfilter python=3.11 cuda-version=12.0
+    cuxfilter python=3.12 cuda-version=12.0
 
 # for CUDA 11.8
 conda install -c rapidsai -c conda-forge -c nvidia \
-    cuxfilter python=3.11 cuda-version=11.8
+    cuxfilter python=3.12 cuda-version=11.8
 ```
 
-Note: cuxfilter is supported only on Linux, and with Python versions 3.10 and 3.11.
+Note: cuxfilter is supported only on Linux, and with Python versions 3.10, 3.11, and 3.12.
 
 ### PyPI
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,6 @@ Please see the [Demo Docker Repository](https://hub.docker.com/r/rapidsai/rapids
 ### CUDA/GPU requirements
 
 - CUDA 11.2+
-- NVIDIA driver 450.80.02+
 - Pascal architecture or better (Compute Capability >=6.0)
 
 ### Conda
@@ -155,9 +154,9 @@ cuxfilter can be installed with conda ([miniconda](https://conda.io/miniconda.ht
 For nightly version `cuxfilter version == 24.10` :
 
 ```bash
-# for CUDA 12.0
+# for CUDA 12.5
 conda install -c rapidsai-nightly -c conda-forge -c nvidia \
-    cuxfilter=24.10 python=3.12 cuda-version=12.0
+    cuxfilter=24.10 python=3.12 cuda-version=12.5
 
 # for CUDA 11.8
 conda install -c rapidsai-nightly -c conda-forge -c nvidia \
@@ -177,6 +176,8 @@ conda install -c rapidsai -c conda-forge -c nvidia \
 ```
 
 Note: cuxfilter is supported only on Linux, and with Python versions 3.10, 3.11, and 3.12.
+
+> Above are sample install snippets for cuxfilter, see the [Get RAPIDS version picker](https://rapids.ai/start.html) for installing the latest `cuxfilter` version.
 
 ### PyPI
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Please see the [Demo Docker Repository](https://hub.docker.com/r/rapidsai/rapids
 ### CUDA/GPU requirements
 
 - CUDA 11.2+
-- Pascal architecture or better (Compute Capability >=6.0)
+- Volta architecture or newer (Compute Capability >=7.0)
 
 ### Conda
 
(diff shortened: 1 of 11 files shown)


notebooks/README.md
```diff
@@ -39,11 +39,11 @@ conda install ipykernel
 
 # for stable rapids version
 conda install -c rapidsai -c numba -c conda-forge -c nvidia \
-    cuxfilter=23.02 python=3.11 cudatoolkit=11.8
+    cuxfilter python=3.12 cudatoolkit=11.8
```
jameslamb, line 42: The Python version change pulled this into the diff, which made me realize the outdated `cuxfilter=23.02` here. I definitely think that should be removed.

Should the `cudatoolkit` also be removed? I'd expect its dependencies like `cuspatial` and `cudf` to manage their own runtime dependencies on things from the CTK, and that separately installing `cudatoolkit` shouldn't be required.

jameslamb:
Now that `cuspatial` has added Python 3.12 support and has nightly packages available (https://github.com/rapidsai/cuspatial/pull/1453), I've restarted CI here.

Will move this out of draft once everything is passing.

conda/recipes/cuxfilter/meta.yaml
```diff
@@ -40,7 +40,6 @@ requirements:
     - numpy >=1.23,<2.0a0
     - packaging
     - panel >=1.0
-    - pyppeteer >=0.2.6
```
jameslamb, line 43: This dependency on `pyppeteer` makes it impossible to install `cuxfilter` in a Python 3.12 environment.

I just pushed https://github.com/rapidsai/cuxfilter/pull/625/commits/7d173772b27cec39e576ce2fa05043194e3b0a12 removing it.

It appears that the latest version of `pyppeteer` ([v1.0.2, published 2.5+ years ago](https://anaconda.org/conda-forge/pyppeteer/files)), puts an upper limit on `websockets<11.0`  and there are not any `websockets` conda packages available that both support Python 3.12 and meet that constraint.

```text
The following packages are incompatible
└─ cuxfilter is installable with the potential options
   ├─ cuxfilter 24.10.00a21 would require
   │  └─ pyppeteer >=0.2.6  with the potential options
   │     ├─ pyppeteer 0.2.6 would require
   │     │  └─ websockets >=9.1,<10.0  with the potential options
   │     │     ├─ websockets [10.0|10.1|10.2|10.3|9.1] would require
   │     │     │  └─ python >=3.7,<3.8.0a0 , which can be installed;
   │     │     ├─ websockets [10.0|10.1|...|9.1] would require
   │     │     │  └─ python >=3.8,<3.9.0a0 , which can be installed;
   │     │     ├─ websockets [10.0|10.1|...|9.1] would require
   │     │     │  └─ python >=3.9,<3.10.0a0 , which can be installed;
   │     │     └─ websockets 9.1 would require
   │     │        └─ python >=3.6,<3.7.0a0 , which can be installed;
   │     └─ pyppeteer 1.0.2 would require
   │        └─ websockets >=10.0,<11.0  with the potential options
   │           ├─ websockets [10.0|10.1|10.2|10.3|10.4] would require
   │           │  └─ python >=3.10,<3.11.0a0 , which can be installed;
   │           ├─ websockets [10.0|10.1|10.2|10.3|9.1], which can be installed (as previously explained);
   │           ├─ websockets [10.0|10.1|...|9.1], which can be installed (as previously explained);
   │           ├─ websockets [10.0|10.1|...|9.1], which can be installed (as previously explained);
   │           └─ websockets 10.4 would require
   │              └─ python >=3.11,<3.12.0a0 , which can be installed;
```

([build link](https://github.com/rapidsai/cuxfilter/actions/runs/10778331677/job/29889656995?pr=625#step:7:395))

I think we can drop the `pyppeteer` dependency here!

* `git grep -i pyppeteer` shows 0 direct uses of it
* it seems the last use of it here was removed when `cuxfilter.assets.screengrab` was removed in #494
* `pyppeteer` appears to be abandoned: https://github.com/pyppeteer/pyppeteer/blob/7dc91ee5173d3836f77800a3774beeaf2b448c0e/README.md#L1

notebooks/README.md
bdice: We should:
- Drop `-c numba` from the line above
- Replace this with something like what we have in cudf:
```
conda install -c rapidsai -c conda-forge -c nvidia \
    cuxfilter=24.10 python=3.12 cuda-version=12.5
```

Make sure `update-version.sh` has a replacement command for versions in `notebooks/README.md`.

jameslamb: Ok great, will push that shortly.

Looks like that's what this project does already do in its main README, so it's just this notebook README that needs to be updated.

https://github.com/rapidsai/cuxfilter/blob/ca63e9d5fdfced7f993f18989ae42409185cff43/README.md#L156-L178

bdice: Maybe we can reduce the content and only list it one place?

jameslamb: yeah even better

jameslamb: Just pushed a commit making these docs in `notebooks/README.md` point back to the installation steps in the top-level `README.md`.

README.md
```diff
@@ -146,7 +145,6 @@ Please see the [Demo Docker Repository](https://hub.docker.com/r/rapidsai/rapids
 ### CUDA/GPU requirements
 
 - CUDA 11.2+
-- NVIDIA driver 450.80.02+
 - Pascal architecture or better (Compute Capability >=6.0)
```
bdice, line 148:
```suggestion
- Volta architecture or newer (Compute Capability >=7.0)
```

README.md
```diff
@@ -156,28 +154,30 @@ cuxfilter can be installed with conda ([miniconda](https://conda.io/miniconda.ht
...
 # for CUDA 11.8
 conda install -c rapidsai-nightly -c conda-forge -c nvidia \
-    cuxfilter=24.10 python=3.11 cuda-version=11.8
+    cuxfilter=24.10 python=3.12 cuda-version=11.8
 ```
 
 For the stable version of `cuxfilter` :
 
 ```bash
 # for CUDA 12.0
```
bdice, line 169:
```suggestion
# for CUDA 12.5
```

README.md
```diff
@@ -156,28 +154,30 @@ cuxfilter can be installed with conda ([miniconda](https://conda.io/miniconda.ht
...
 conda install -c rapidsai -c conda-forge -c nvidia \
-    cuxfilter python=3.11 cuda-version=12.0
+    cuxfilter python=3.12 cuda-version=12.0
```
bdice, line 171:
```suggestion
    cuxfilter python=3.12 cuda-version=12.5
```

bdice (commented):
A few related cleanups -- I'll apply these and trigger a `/merge`.

bdice (approved):

PR #627, opened 2024-09-17
Update to flake8 7.1.1.

bdice:
We need to update flake8 to fix a false-positive that appears with older flake8 versions on Python 3.12.

### DIFF
--- a/python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py
+++ b/python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py
@@ -21,7 +21,6 @@ class BaseStackedLine(BaseChart):
     y_range: Tuple = None
     use_data_tiles = False
     y: list = []
-    colors: list = []
     default_colors = ["#8735fb"]
     box_selected_range = None
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
                 # Explicitly specify the pyproject.toml at the repo root, not per-project.
                 args: ["--config", "pyproject.toml"]
       - repo: https://github.com/PyCQA/flake8
-        rev: 6.0.0
+        rev: 7.1.1
         hooks:
               - id: flake8
                 args: ["--config=.flake8"]
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -7,6 +7,7 @@ channels:
 - nvidia
 dependencies:
 - bokeh>=3.1
+- bokeh_sampledata
 - cuda-version=11.8
 - cudatoolkit
 - cudf==24.10.*,>=0.0.0a0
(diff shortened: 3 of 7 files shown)


python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py
```diff
@@ -21,7 +21,6 @@ class BaseStackedLine(BaseChart):
     y_range: Tuple = None
     use_data_tiles = False
     y: list = []
-    colors: list = []
```
jameslamb, line 24: This one doesn't look right to me. I don't see `self.colors` getting set in `__init__()`.

https://github.com/rapidsai/cuxfilter/blob/53537fe4fb2dff70789d8cfb81d2c058adc6488c/python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py#L59

And I found at least one place that appears to rely on this attribute being defined on an instance:

https://github.com/rapidsai/cuxfilter/blob/53537fe4fb2dff70789d8cfb81d2c058adc6488c/python/cuxfilter/charts/datashader/plots.py#L384

https://github.com/rapidsai/cuxfilter/blob/53537fe4fb2dff70789d8cfb81d2c058adc6488c/python/cuxfilter/charts/datashader/plots.py#L394

What was the the `flake8` error that this change addresses? Maybe there's another way we could address it, like a more specific type hint in the initialization of `colors`.

jameslamb (changes requested):
Support updating `flake8`, but I left one comment... I'm concerned about the change to `BaseStackedLine`.

python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py
bdice: It's defined as a property here: https://github.com/rapidsai/cuxfilter/blob/f9a617d889bd01d45687e16f96beebc06f6acdfd/python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py#L52-L56

flake8 was mad that this object had both an attribute and a property with the same name, and it seemed to me that the property was the correct choice to keep. What do you think?

jameslamb: ahhhh my `git grep -E '\.colors`' wouldn't have caught it being defined with `@property`, should have thought of that!

Ok then yeah this is totally fine, and I agree with preferring the `@property` one. Thanks for explaining that.

jameslamb (approved):

python/cuxfilter/charts/core/non_aggregate/core_stacked_line.py
bdice: cc: @AjayThorve for awareness

AjayThorve: thanks, this seems correct, and yes property is the right one to keep

AjayThorve (approved):
LGTM, thanks!

AjayThorve:
seems like the CI is failing, due to a new dependency introduced to `bokeh_sampledata`. Would need to add it to `dependencies.yaml` to fix the issue for notebooks and docs

PR #665, opened 2025-02-21
require sphinx<8.2.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/155

Fixes docs builds by temporarily putting a ceiling on `sphinx`.


bdice (approved):
We might want to use a lower bound of `>8` to match other repos.

jameslamb:
> We might want to use a lower bound of `>8` to match other repos.

Sure, that's fine. Just pushed https://github.com/rapidsai/cuxfilter/pull/665/commits/85256b5789d1ffd3ec6d9e498e5037a5d03ad8ea with that change.

PR #669, opened 2025-04-02
Port all conda recipes to `rattler-build`

gforsyth:
Port all condabuild recipes over to use `rattler-build` instead.

Contributes to rapidsai/build-planning#47

- To satisfy `rattler-build`, this changes all the licenses in the `pyproject.toml` files to the SPDX-compliant `Apache-2.0` instead of `Apache 2.0`


conda/recipes/cuxfilter/recipe.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - setuptools
+  run:
+    - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+    - cudf =${{ minor_version }}
+    - cupy >=12.0.0
+    - cuspatial =${{ minor_version }}
+    - dask-cudf =${{ minor_version }}
+    - datashader >=0.15
+    - geopandas >=0.11.0
+    # TODO: shapely 2.1.0 transitive dependency breaks geopandas
```
bdice, line 41: Is this still true, or has it been solved? I think someone said a new geopandas version fixed this.

bdice (approved):
Approved with one question.

conda/recipes/cuxfilter/recipe.yaml
gforsyth: heh, i think that was me.  I'll try removing it and see what happens

PR #670, opened 2025-04-03
Fix recently introduced docs build issue

AjayThorve:
Seems like the newer version of bokeh post 3.6.3 is casuing issues with jupyter_sphinx cell execution. This PR adds a pin to last known working version.


gforsyth:
There was a release of Shapely about 8 hours ago and 2.1.0 is apparently a "major" release.  So we should also pin to <2.1.0 for now (as the deprecated usage of `shapely.geos` is via `geopandas`)

gforsyth:
Resolved via #672

PR #675, opened 2025-04-09
fix(docs): pin bokeh upper-bound for all docs environments

gforsyth:
Missed a few environments that led to further failures in the docs builds.


gforsyth:
xref #672 #670

jakirkham:
Would have thought we need to run `rapids-dependency-file-generator`, but it looks like these pins are already there

https://github.com/rapidsai/cuxfilter/blob/7cc2a49e7c32bad83ef11677a5b57f3f09229400/python/pyproject.toml#L21

https://github.com/rapidsai/cuxfilter/blob/7cc2a49e7c32bad83ef11677a5b57f3f09229400/conda/environments/all_cuda-128_arch-x86_64.yaml#L9

Looks like the Conda recipe already has the pin too

https://github.com/rapidsai/cuxfilter/blob/7cc2a49e7c32bad83ef11677a5b57f3f09229400/conda/recipes/cuxfilter/meta.yaml#L29

jakirkham (approved):

jakirkham:
Thanks Gil! 🙏

Maybe we should double check where the RAPIDS dependency file generator style check is picking up this kind of thing. Want to make sure this issue is caught in CI in the future

PR #681, opened 2025-05-02
Replace cuspatial

### LINKED ISSUE #678: [FEA] Remove cuspatial dependency for point-in-polygon in cuxfilter
Since cuspatial is being discontinued, We'd like to remove the dependency on cuspatial for the point-in-polygon method used in cuxfilter, specifically for lasso selection.

The current use case involves checking many points against a single polygon (e.g., a user-drawn lasso), which is a significantly simpler scenario than handling many points vs. many polygons.

A custom solution using a Numba kernel should be sufficient for this case. The point-in-polygon algorithm (e.g., ray casting or winding number) is well-known and can be efficiently implemented for the single-polygon scenario without relying on a full spatial library like cuspatial.

Next Steps:

Prototype a Numba-based point-in-polygon kernel.

Benchmark against cuspatial for performance comparison.

Integrate and test within cuxfilter's selection logic.

Cc @exactlyallan @bdice

AjayThorve:
Since cuspatial is being discontinued, We'd like to remove the dependency on cuspatial for the point-in-polygon method used in cuxfilter, specifically for lasso selection.

This PR fixes #678.

cc @jakirkham @bdice


python/cuxfilter/charts/core/non_aggregate/utils.py
```diff
@@ -1,13 +1,177 @@
...
+            raise ValueError("Polygon must have at least 3 vertices.")
+
+    except (ValueError, TypeError):
+        # Handle invalid polygon formats or insufficient vertices
+        # Return all False as per the check at the start
+        if isinstance(df.index, cudf.MultiIndex):
+            idx = cudf.RangeIndex(len(df))
+        else:
+            idx = df.index
+        return cudf.Series([False] * len(df), index=idx, dtype="bool")
```
bdice, line 153: This requires a host allocation of size `len(df)` which can be avoided. See cuSpatial: https://github.com/rapidsai/cuspatial/blob/fc344cec429c6057917d22debe52083960df616d/python/cuspatial/cuspatial/utils/binpred_utils.py#L23-L25

You can see a microbenchmark here: https://colab.research.google.com/gist/bdice/b3d67dc7dd1a3055f52312cc1168e91f/avoid_host_allocation.ipynb

Avoiding a host allocation & copy gives a huge speedup.
```suggestion
        return cudf.Series(cp.zeros(len(df), dtype=cp.bool_), index=idx)
```

python/cuxfilter/charts/core/non_aggregate/utils.py
```diff
@@ -1,13 +1,177 @@
...
+
+    # 3. Prepare output array
+    out_gpu = cp.zeros(num_points, dtype=cp.bool_)
+
+    # 4. Configure and launch kernel
+    threads_per_block = 128
```
bdice, line 159: Naively, 256 can be a better default here. Depends on the kernel. Please check whether performance is affected.

python/cuxfilter/tests/charts/core/non_aggregate/test_utils.py
```diff
@@ -0,0 +1,122 @@
...
+            0.5,
+            1.5,
+            0.5,
+            1.5,
+            0.0,
+            2.0,
+            1.0,
+            1.0,
+        ],
+        # Inside, Inside, Outside, Outside, Boundary, Boundary,
```
bdice, line 24: Be very cautious with boundary testing, as numerical precision issues can cause the behavior to differ in unstable ways, causing tests to be flaky as versions/platforms change. In my understanding, a user will be doodling a polygon on a map and the boundary behavior won't matter so precisely. I would advise against doing any boundary testing here.

bdice:
@AjayThorve Please also remove cuSpatial references from these files:
- https://github.com/rapidsai/cuxfilter/blob/branch-25.06/notebooks/RAPIDS%20Visualization%20Guide/RAPIDS%20Viz%20Guide%20Notebook.ipynb
- https://github.com/rapidsai/cuxfilter/blob/33621a4054479d65156421f27b55021ad8f98a1f/ci/release/update-version.sh#L48
- https://github.com/rapidsai/cuxfilter/blob/33621a4054479d65156421f27b55021ad8f98a1f/notebooks/RAPIDS%20Visualization%20Guide/README.md?plain=1#L8
- https://github.com/rapidsai/cuxfilter/blob/33621a4054479d65156421f27b55021ad8f98a1f/conda/recipes/cuxfilter/recipe.yaml#L37
- https://github.com/rapidsai/cuxfilter/blob/33621a4054479d65156421f27b55021ad8f98a1f/notebooks/RAPIDS%20Visualization%20Guide/RAPIDS%20Viz%20Catalog.ipynb#L49

... and grep for anything else, please!

python/cuxfilter/charts/core/non_aggregate/utils.py
```diff
@@ -1,13 +1,177 @@
...
+    ----------
+    df : cudf.DataFrame
+        DataFrame containing the points.
+    x : str
+        Column name for the x-coordinates of the points.
+    y : str
+        Column name for the y-coordinates of the points.
+    polygon_coords : list of tuples or similar array-like
+        Coordinates of the polygon vertices, e.g., [(x1, y1), (x2, y2), ...].
+        The polygon should be closed.
```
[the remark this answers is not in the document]
AjayThorve: Yes, closed here in this context means first and last vertex coordinates must be equal. The way lasso selection is setup in cuxfilter, the polygon will always be closed. 

output dashboard

python/cuxfilter/tests/charts/core/non_aggregate/test_utils.py
AjayThorve: that's a fair point, I agree with this

python/cuxfilter/charts/core/non_aggregate/utils.py
```diff
@@ -1,13 +1,177 @@
...
+from numba.core.errors import NumbaPerformanceWarning
+import cupy as cp
+import cudf
+import warnings
+
+
+# Define the CUDA kernel using numba.cuda.jit
+# This kernel implements the ray casting algorithm
+@cuda.jit(device=True)
+def point_in_polygon_ray_cast(px, py, polygon_xy, num_vertices):
```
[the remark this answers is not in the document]
AjayThorve: figure 8 works fine: 
image

concave/convex work fine as well:
image

image

python/cuxfilter/charts/core/non_aggregate/utils.py
AjayThorve: Not much, I am getting similar results eitherways:

fluctuates between multiple runs:

--- Calling point_in_polygon (1st time - includes compilation) ---
First call duration: 15-25ms

--- Calling point_in_polygon (2nd time - uses compiled kernel) ---
Second call duration: 5-6ms

python/cuxfilter/charts/core/non_aggregate/utils.py
AjayThorve: Will update it, thanks!

python/cuxfilter/charts/core/non_aggregate/utils.py
AjayThorve: updated to 256 nonetheless

bdice (approved):
Amazing! Thank you so much.

jakirkham:
Thanks all! 🙏

PR #700, opened 2025-06-17
Drop unneeded cuSpatial dependency

jakirkham:
Previously cuSpatial was a dependency of cuxfilter. However the logic was rewritten to drop cuSpatial and drop the dependency in PR ( https://github.com/rapidsai/cuxfilter/pull/681 ) as part of 25.06. However a few cuSpatial references were either missed or snuck back in 25.08. So this cleans them up


AjayThorve (approved):
LGTM, thanks!

bdice (approved):

PR #702, opened 2025-06-25
Remove nvidia and dask channels

vyasr:
Now that we have dropped support for CUDA 11 we no longer require the nvidia channel.
With the changes in https://github.com/rapidsai/rapids-dask-dependency/pull/85, RAPIDS now only uses released versions of dask, so we no longer need the dask channel either.
Contributes to https://github.com/rapidsai/build-planning/issues/184


jakirkham:
Thanks Vyas! 🙏

Do we want to update the docs as well? Or save that for a separate pass?

https://github.com/rapidsai/cuxfilter/blob/9ce9a15810f2cc14e19f894d2937e24ae888d68c/README.md?plain=1#L158

vyasr:
> Thanks Vyas! 🙏
> 
> Do we want to update the docs as well? Or save that for a separate pass?
> 
> https://github.com/rapidsai/cuxfilter/blob/9ce9a15810f2cc14e19f894d2937e24ae888d68c/README.md?plain=1#L158

Good idea, done.

vyasr:
@jakirkham sorry bit of a race condition here. I pushed a couple of extra changes to remove CUDA 11 references that I noticed. Would you mind taking another quick pass?

jakirkham (approved):
Reapproving

KyleFromNVIDIA (approved):

PR #726, opened 2025-09-24
Update `RAPIDS_BRANCH`, codify changes in `update-version.sh`

KyleFromNVIDIA:
This PR updates `RAPIDS_BRANCH` and modifies `update-version.sh` to make future changes automatically.


bdice (changes requested):
This repo doesn't use rapids-cmake. Should we close?

PR #781, opened 2026-05-11
Constrain NumberChart expressions

exactlyallan:
## Summary

This tightens `NumberChart` expression handling so supported expressions are evaluated through a small, explicit dataframe-column arithmetic path.

## Changes

- Add a constrained expression evaluator for `NumberChart`.
- Use it in both chart generation and reload paths.
- Add focused coverage for supported arithmetic and unsupported expression forms.

## Validation

```bash
PYTHONPYCACHEPREFIX=/private/tmp/cuxfilter-pycache \
  python3 -m py_compile \
  python/cuxfilter/charts/panel_widgets/plots.py \
  python/cuxfilter/tests/charts/panel_widgets/test_plots.py

git diff --check HEAD~1 HEAD
```

Both local checks passed.

Full test and lint suites were not run locally because this non-GPU Mac checkout does not have the RAPIDS runtime/test environment installed. This PR should use the project CI for full validation.


bdice (approved):
In some cases you can also use https://docs.python.org/3/library/ast.html#ast.literal_eval, not sure if that would be useful here since this appears to be acting on Series objects.

jakirkham:
Thanks Bradley! 🙏

Think that is mainly used for constants like...

```python
ast.literal_eval("2")
ast.literal_eval("2.3")
```

However it doesn't handle expressions like...

```python
ast.literal_eval("2.2 + 1")    # raises ValueError
```

This lines up with [this section from the docstring]( https://docs.python.org/3/library/ast.html#ast.literal_eval ):

> This can be used for evaluating strings containing Python values without the need to parse the values oneself. It is not capable of evaluating arbitrarily complex expressions, for example involving operators or indexing.

So think we are back to rolling our own expression handling as Allan has done. Though it would certainly be nice if there was a simple and safe way to do this. For now maybe we can stick with what we have here.

jakirkham (approved):
Thanks Allan! 🙏

PR #789, opened 2026-06-01
Remove numba-cuda upper bound

brandon-b-miller:
This PR removes the upper bound on the `numba-cuda` dependency,
leaving only the existing lower bound pins in place.

<hr>

Part of this work:

* https://github.com/rapidsai/build-planning/issues/245


jakirkham:
Thanks Brandon! 🙏

cuxfilter is being sunset: https://github.com/rapidsai/docs/pull/786

So think we can close this

PR #793, opened 2026-06-10
update development status, remove branch CI

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/291

RSN 60 announces that there won't be any more `cuxfilter` packages published after 26.06 (https://docs.rapids.ai/notices/rsn0060/).

From my offline conversation with @exactlyallan , it sounds like the repo may be repurposed (details TBD) so shouldn't yet be archived.

So this proposes the following on `main`:

* stop running CI on merges to `main`
* preserve PR CI for now
* add a note to the README pointing to the RSN and explaining the current status
* pin install commands to 26.06

While doing this, I'm also taking the opportunity to remove some outdated stuff:

* the "external testing" setup
* an outdated note about using the `rapidsai/base` container images


gforsyth (approved):

jameslamb:
one more cuDF API removal for the road 🥹 

```text
File /opt/conda/envs/docs/lib/python3.14/site-packages/holoviews/core/data/cudf.py:141, in cuDFInterface.values(cls, dataset, dim, expanded, flat, compute, keep_index)
    139     return data
    140 elif compute:
--> 141     return data.values_host
    142 try:
    143     return data.values

AttributeError: 'Series' object has no attribute 'values_host'
```

([build link](https://github.com/rapidsai/cuxfilter/actions/runs/27283500096/job/80585377250?pr=793))

Looks like the primary cause is https://github.com/rapidsai/cudf/pull/22649

Notice that the errors' coming from `holoviews` code.

It was fixed in https://github.com/holoviz/holoviews/pull/6852, but that hasn't made it into a release yet (it's in the `v1.23.0rc1` tag, so I guess a 1.23.0 release will have it).

We're about to stop development here so can just put a ceiling on the `cudf` dependency. And I'll report this to cuDF maintainers in case they want to do something for it.

conda/recipes/cuxfilter/recipe.yaml
```diff
@@ -34,9 +33,9 @@ requirements:
   run:
     - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
     - certifi >=2026.1.4
-    - cudf =${{ minor_version }}
+    - cudf =26.6
```
jameslamb, line 36: Proposing pinning to the last stable release of `cudf` to address https://github.com/rapidsai/cuxfilter/pull/793#issuecomment-4684889837 and protect CI here from future breaking changes in `cudf`.

There won't be a `cuxfilter` 26.08, I think it's appropriate for it to pin against the last releases of RAPIDS libraries it was tested against for now.

(this CI may go away completely anyway depending on how the repo changes in the near future)

gforsyth (approved):
:shipit: 

I agree with pinning deps and workflows to 26.06

